PR #39, opened 2024-08-28
revise: implement membership cancellation mechanism (TS-461)

maikelordaz:
Here also applied defaulting a member as of previous conversations, done inside here because I'm using the `Defaulted` state for the cancellation implementation.

PS: While I'm writing this I saw the word `cancellation` with two `L` but in the contract I wrote `Canceled` I looked in the web and I find that `Canceled` and `Cancelled` are both correct, so let me know which one to use


contracts/takasure/TakasurePool.sol
```diff
@@ -349,51 +357,94 @@ contract TakasurePool is
...
-            currentTimestamp > lastPaidYearStartDate + year + gracePeriod ||
-            currentTimestamp > membershipStartTime + membershipDuration
-        ) {
+        if (currentTimestamp > firstLimitTimestamp) {
             revert TakasureErrors.TakasurePool__InvalidDate();
+        } else if (currentTimestamp >= secondLimitTimestamp) {
+            // Update the state, this will allow to cancel the membership
+            reserve.members[msg.sender].memberState = MemberState.Defaulted;
+
+            emit TakasureEvents.OnMemberDefaulted(reserve.members[msg.sender].memberId, msg.sender);
```
elallamTaka, line 399: Hey Maikel, i can see that if a user tries to make a payment after his grace days passing, then his state is updated to defaulted. This behavior is out of the concern of this function. Meaning, `recurringPayment ` must only focus on making the payment, and failing if its the wrong user state. Thus here you can simply return an error such as `paymentTooLate`. It will be the responsibility of the CS to mark the user as defaulted, or the responsibility of a cron job.

contracts/takasure/TakasurePool.sol
```diff
@@ -349,51 +357,94 @@ contract TakasurePool is
...
+
+    /**
+     * @notice Method to cancel a membership
+     * @dev To be called by anyone
+     */
+    function cancelMembership(address memberWallet) external notZeroAddress(memberWallet) {
+        _cancelMembership(memberWallet);
+    }
+
     function recurringPayment() external {
```
elallamTaka, line 376: Since this is an action the user takes, I suggest we follow the same naming convention of starting with the verb like what we did for `payContribution`. Maybe `payRecurringContribution` ?

contracts/takasure/TakasurePool.sol
```diff
@@ -349,51 +357,94 @@ contract TakasurePool is
...
+                msg.sender,
+                true
+            );
+
+            emit TakasureEvents.OnRecurringPayment(
+                msg.sender,
+                reserve.members[msg.sender].memberId,
+                reserve.members[msg.sender].lastPaidYearStartDate,
+                reserve.members[msg.sender].totalContributions,
+                reserve.members[msg.sender].totalServiceFee
```
elallamTaka, line 426: I dont think the total contribution here is required to be emitted. Rather, the frontend wants to display the contribution paid as a transaction so perhaps we can emit the same values emitted for the first payment.

contracts/takasure/TakasurePool.sol
```diff
@@ -349,51 +357,94 @@ contract TakasurePool is
...
-        emit TakasureEvents.OnRecurringPayment(
-            msg.sender,
-            reserve.members[msg.sender].memberId,
-            reserve.members[msg.sender].lastPaidYearStartDate,
-            reserve.members[msg.sender].totalContributions,
-            reserve.members[msg.sender].totalServiceFee
-        );
+        uint256 currentTimestamp = block.timestamp;
+        uint256 lastPaidYearStartDate = member.lastPaidYearStartDate;
+        uint256 limitTimestamp = member.membershipStartTime + lastPaidYearStartDate + (30 days); // 30 days after the new membership year
```
elallamTaka, line 438: This is a default, which means users default if they dont make the payment before the end of their membership year, so we must remove 30 days from here. 

To recap, if a user misses a payment, he defaults. If he fails to pay within 30 days (grace period) his membership gets cancelled.

contracts/takasure/TakasurePool.sol
```diff
@@ -625,6 +676,28 @@ contract TakasurePool is
         );
     }
 
+    function _cancelMembership(
```
elallamTaka, line 679: The cancellation is missing running some key algorithms such as distributing the surplus according to the cancellation year (based on the reqs). However I prefer we update it later after we do the optimization. I have created a backlog ticket for it.

elallamTaka (changes requested):

contracts/takasure/TakasurePool.sol
maikelordaz: Commit e307311

contracts/takasure/TakasurePool.sol
maikelordaz: Commit 7d78560

contracts/takasure/TakasurePool.sol
maikelordaz: Commit c18de38

contracts/takasure/TakasurePool.sol
maikelordaz: Commit a4d5b2c

contracts/takasure/TakasurePool.sol
maikelordaz: Got it

elallamTaka (approved):

PR #50, opened 2024-11-07
revise: Fix refund process (TS-923)

maikelordaz:
Fix for the refunds as discussed. The refund will reduce first the currentAmount, then if needed the referralReserve, then the toRepool and if it is not enough it will revert with a custom error that includes the amount to send to the contract

=================================================================

A peacefull image as a celebration to move on with other parts of the work. Well done!

Green Lagoon, Humbold Peak 02/15/2010
P1010252


maikelordaz:
@elallamTaka the fix is done. I move the error to revert earlier to save some gas, and the main issue was inn the payContribution function

elallamTaka (approved):
Looks good! My only optional suggestion is to define something like `tDAO storage DAOData = nameToDAOData[_tDAOName];` in the begining of the refund function. Then access each object like so `DAOdata.currentAmount. `

I think its more gas optimized since memory access is cheaper than storage access?

elallamTaka (changes requested):
Update the refund to do the following: 
* For feeToOperator/ discount take from referral pool/ repool/ operator (how much the op topped up is basically balanceOf(address) - currentAmount - toRepool - referalPool. 
* For contributionAfterFee, take from contribution pool

For each users refund, deduct the amount according to how they were deposited, most importantly, take the contributionAfterFee from currentAmount.

PR #58, opened 2025-01-28
revise: ccip pay contribution (TS - 1188)

maikelordaz:
This PR allow to pay contributions from different chains.

<h2>Important changes since CCIP template</h2>

- I changed the contracts names, now they are `TLDCcipReceiver` and `TLDCcipSender`
- Now `TLDCcipSender` handle the CCIP fees either in LINK or native token, although I think the best would be to choose one. Remember this have to be kept in the contract and it is not the contract call fee, it is the fee for the use of the CCIP
- I changed the main function name to `payContribution`
- As the contracts are not upgradeables, and it can be that in the future we accept more chains and/or tokens, now there are whitelists for tokens and chains

<h2>About the payment process</h2>

- It will call the `payContributionOnBehalfOf` instead of the `payContribution` one. This to be able to use the coupon codes
- In the sender contract, there are security checks for the coupon, if it is higher than 0, then the caller must be the backend
- In the `ReferralGateway` the caller must be either the caller or the receiver contract

<h2>Limitations</h2>

As the CCIP only allow certain tokens transfers, the USDC that can be used is the one deployed by circle in testnets, and the usdc in the referralGateway is the one deployed by me. There are two workarounds for this

1. Call `setUsdcAddress` and change it to the one deployed by circle. This means that every test made on uat will fail, this because get this test tokens is easy, but not easy to get enough
2. This is the one I used, I just add a line to handle the transfer, if the caller is the receiver contract, this will allow to continue testing in uat

Yo can see the changes in this commented lines.

```solidity
            if (amountToTransfer > 0) {
                if (msg.sender == ccipReceiverContract) {
                    usdc.safeTransferFrom(ccipReceiverContract, address(this), amountToTransfer);

                    // Note: This is a temporary solution to test the CCIP integration in the testnet
                    // This is because in testnet we are using a different USDC contract for easier testing
                    // IERC20(0x75faf114eafb1BDbe2F0316DF893fd58CE46AA4d).safeTransferFrom(
                    //     ccipReceiverContract,
                    //     address(this),
                    //     amountToTransfer
                    // );
                } else {
                    usdc.safeTransferFrom(_newMember, address(this), amountToTransfer);
                }
            }
``` 

<h2>Deployed contracts</h2>

Due to limitations with the native test tokens I deployed only in:

- [Ethereum Sepolia](https://sepolia.etherscan.io/address/0x9A44C15fE0bdAb9b5487269BEBF49540269250eE)
- [Base Sepolia](https://sepolia.basescan.org/address/0x0680d25eFFaAC6D8ed599B4211bd0C20273748bC)
- [Optimism Sepolia](https://sepolia-optimism.etherscan.io/address/0x7A156Dd3ABaD4Be84952877277E302491A703D0e)

<h2>Test steps</h2>

As a recomendation to test, please follow this steps
- Use the Ethereum Sepolia one
- Ask for test ETH in this [faucet](https://cloud.google.com/application/web3/faucet)
- Ask for test USDC in this [faucet](https://faucet.circle.com/?_gl=1*1nh51fu*_gcl_au*MTk1OTk2NTIxNS4xNzM0OTA5MTcx*_ga*MjAwMjk4NDIzMy4xNzM0OTA5MTcx*_ga_GJDVPCQNRV*MTczNjE3ODgxMS40LjAuMTczNjE3ODgxMy41OC4wLjA.)
- Tip for the test USDC, you can ask for 10 tokens each hour, but there is no limitation if you ask for other 10 in another address and then you put them all together in the same account
- Remember to have at least 25 test USDC, if not the function will failed
- In etherscan go to the usdc in Ethereum Sepolia and approve the Sender contract to spend your tokens, this is the [address](https://sepolia.etherscan.io/address/0x1c7D4B196Cb0C7B01d743Fbc6116a902379C7238)
- In etherscan go to the Sender contract and call the `payContribution` function with the next values
`amountToTransfer` will be the amount of test USDC to transfer, `tokenToTransfer` the USDC Ethereum Sepolia address, `contribution` the contribution to pay, `tDAOName` use `The Life DAO test_2`, `parent` if you have a parent address you can use it here, if not use the address 0, `couponAmount` if you are not using the backend address this must be 0, if not it will fail, `feesInLink` must be true, the contracts have enough LINK balance. 
- Remember, if `couponAmount` is 0 then `contribution` is equal to `amountToTransfer` 
- Take the tx hash and go to the [CCIP explorer](https://ccip.chain.link/) and paste here your tx hash
- Wait
- Wait a bit longer
- Maybe wait a little more
- After maybe 20-30 minutes it should be done
- You'll see the Arbitrum Sepolia tx hash and can inspect it, you'll find a lot of events related to the transfer and mint of USDC, events from the CCIP and our events


contracts/chainlink/ccip/TLDCcipReceiver.sol
```diff
@@ -75,25 +81,59 @@ contract Receiver is CCIPReceiver, Ownable2Step {
         address _usdc,
         address _referralGateway
     ) CCIPReceiver(_router) Ownable(msg.sender) {
-        require(_usdc != address(0), Receiver__InvalidUsdcToken());
+        require(_usdc != address(0), TLDCcipReceiver__InvalidUsdcToken());
         usdc = IERC20(_usdc);
         referralGateway = _referralGateway;
```
elallamTaka, line 86: Since the referralGateway will be part of the entry module, I suggest we rename the input of this function to something more generic? Like in the description i can see you are saying "staker contract". How about protocolGateway which kind of indicates the entry of any protocol. 

Also maybe a function to change the staker contract ?

contracts/chainlink/ccip/TLDCcipReceiver.sol
```diff
@@ -49,19 +50,24 @@ contract Receiver is CCIPReceiver, Ownable2Step {
...
+        require(
+            _sourceChainSelector != 0 &&
+                isSourceChainAllowed[_sourceChainSelector] &&
+                isSenderAllowed[_sender],
+            TLDCcipReceiver__NotAllowedSource()
+        );
         _;
     }
 
     modifier onlySelf() {
```
elallamTaka, line 69: is there an existing modifier like this one in the ownable lib by openzep ?

contracts/chainlink/ccip/TLDCcipReceiver.sol
```diff
@@ -114,17 +163,20 @@ contract Receiver is CCIPReceiver, Ownable2Step {
...
-            Receiver__MessageNotFailed(messageId)
+            TLDCcipReceiver__MessageNotFailed(messageId)
         );
         failedMessages.set(messageId, uint256(ErrorCode.RESOLVED));
 
         Client.Any2EVMMessage memory message = messageContentsById[messageId];
 
+        // Low level call to the referral gateway
+        (bool success, ) = referralGateway.call(message.data);
+        require(success, TLDCcipReceiver__CallFailed());
```
elallamTaka, line 178: What if the retry fails as well? I think there must be a way to somehow return the money to the user who originally sent it, meaning transfer it to his address on arb chain so he can try to pay directly through interface. What do you think?

contracts/chainlink/ccip/TLDCcipSender.sol
```diff
@@ -0,0 +1,320 @@
...
+
+    modifier notZeroAddress(address addressToCheck) {
+        require(addressToCheck != address(0), TLDCcipSender__AddressZeroNotAllowed());
+        _;
+    }
+
+    /**
+     * @param _router The address of the router contract.
+     * @param _link The address of the link contract.
+     * @param _receiverContract The address of the referral contract. This will be the only receiver
```
elallamTaka, line 63: address of referral contract? i thought the address of the tld receiver must be passed here

contracts/chainlink/ccip/TLDCcipSender.sol
```diff
@@ -0,0 +1,320 @@
...
+     * @param _owner admin address
+     */
+    constructor(
+        address _router,
+        address _link,
+        address _receiverContract,
+        uint64 _chainSelector,
+        address _owner,
+        address _backendProvider
+    ) Ownable(_owner) {
```
elallamTaka, line 74: Why is it that the owner here is passed, but in the receiver the owner is msg.sender?

contracts/chainlink/ccip/TLDCcipSender.sol
```diff
@@ -0,0 +1,320 @@
...
+        router = IRouterClient(_router);
+        linkToken = IERC20(_link);
+        receiverContract = _receiverContract;
+        destinationChainSelector = _chainSelector;
+        backendProvider = _backendProvider;
```
elallamTaka, line 79: setting the backend provider here means that everytime we update the backend provider in the referral, then we mut update it here. Is there a way to read the backend provider role address from here so we dont need to update it in several contracts?

contracts/chainlink/ccip/TLDCcipSender.sol
```diff
@@ -0,0 +1,320 @@
...
+        Client.EVMTokenAmount memory tokenAmount = Client.EVMTokenAmount({
+            token: _token,
+            amount: _amount
+        });
+        tokenAmounts[0] = tokenAmount;
+
+        // Function to call in the receiver contract
+        // payContributionOnBehalfOf(uint256 contribution, string calldata tDAOName, address parent, address newMember, uint256 couponAmount)
+        bytes memory dataToSend = abi.encodeWithSignature(
+            "payContributionOnBehalfOf(uint256,string,address,address,uint256)",
```
elallamTaka, line 299: so cccip will always use pay on behalf right?

contracts/chainlink/ccip/TLDCcipSender.sol
```diff
@@ -0,0 +1,320 @@
...
+            _couponAmount
+        );
+
+        // EVM2AnyMessage struct
+        Client.EVM2AnyMessage memory message = Client.EVM2AnyMessage({
+            receiver: abi.encode(receiverContract),
+            data: dataToSend,
+            tokenAmounts: tokenAmounts,
+            extraArgs: Client._argsToBytes(
+                Client.EVMExtraArgsV2({gasLimit: 1_000_000, allowOutOfOrderExecution: true})
```
elallamTaka, line 313: is it safe to have these hardcoded like this?

contracts/chainlink/ccip/TLDCcipSender.sol
```diff
@@ -0,0 +1,320 @@
...
+    function payContribution(
+        uint256 amountToTransfer,
+        address tokenToTransfer,
+        uint256 contribution,
+        string calldata tDAOName,
+        address parent,
+        uint256 couponAmount,
+        bool feesInLink
+    ) external returns (bytes32 messageId) {
+        require(isSupportedToken[tokenToTransfer], TLDCcipSender__NotSupportedToken());
```
elallamTaka, line 139: To confirm, the only supported token is usdc now, so this will fail for others. And if we want to add another token, then we either enable it in the pool, or somehow implement a swap in between?

contracts/chainlink/ccip/TLDCcipSender.sol
```diff
@@ -0,0 +1,320 @@
...
+     * @param feesInLink If true, the fees will be paid in LINK tokens. If false, the fees will be paid in native gas.
+     * @return messageId The ID of the message that was sent.
+     */
+    function payContribution(
+        uint256 amountToTransfer,
+        address tokenToTransfer,
+        uint256 contribution,
+        string calldata tDAOName,
+        address parent,
+        uint256 couponAmount,
```
elallamTaka, line 136: If we construct the message from these inputs, then its possible that we will need to upgrade this contract if the inputs change in the referral contract. Why not make this upgradable? (not that I prefer). Otherwise it might be safer to have the backend / frontend construct the calldata

contracts/chainlink/ccip/TLDCcipSender.sol
```diff
@@ -0,0 +1,320 @@
...
+     * @param amountToTransfer token amount to transfer to the receiver contract in the destination chain.
+     * @param tokenToTransfer The address of the token to be transferred. Must be in the list of supported tokens.
+     * @param contribution The amount of the contribution to be paid in the TLD contract.
+     * @param tDAOName The name of the DAO to point in the TLD contract.
+     * @param parent The address of the parent if the caller has a referral.
+     * @param couponAmount The amount of the coupon if the caller has one.
+     * @param feesInLink If true, the fees will be paid in LINK tokens. If false, the fees will be paid in native gas.
+     * @return messageId The ID of the message that was sent.
+     */
+    function payContribution(
```
elallamTaka, line 130: My preference is to keep this `sendMessage` for naming conventions

contracts/referrals/ReferralGateway.sol
```diff
@@ -160,6 +165,7 @@ contract ReferralGateway is
     error ReferralGateway__NotKYCed();
     error ReferralGateway__tDAONotReadyYet();
     error ReferralGateway__NotEnoughFunds(uint256 amountToRefund, uint256 neededAmount);
+    error ReferralGateway__WrongCaller();
```
elallamTaka, line 168: not authorized caller ?

elallamTaka (changes requested):

contracts/chainlink/ccip/TLDCcipReceiver.sol
maikelordaz: Like `onlySelf`? never seen it, but I'll check

contracts/chainlink/ccip/TLDCcipSender.sol
maikelordaz: Yes, wrong writing, change it in commit 31ac54c

contracts/chainlink/ccip/TLDCcipSender.sol
maikelordaz: This is for the deterministic deployment, for that I need an address, without activity in any address, so the nonce is the same in every chain, that way it can be deployed to the same address in every chain, but that address can not be used for anything else, because that will mess with the nonce. 
Have not done it because the Create2 deployer builtin in foundry scripts also use the hash of the constructor arguments, and as in here, the router address will change, then it wont work. So I have either to find another Create2 deployer or create the deployer myself and deploy it in each network, dont think thats a priority right now so I'll leave that for now

contracts/chainlink/ccip/TLDCcipSender.sol
maikelordaz: Yes, because of two reasons:
1. That way the coupons feature can be used
2. `payContribution` use the `msg.sender` and that one is the Receiver contract

contracts/chainlink/ccip/TLDCcipSender.sol
maikelordaz: According to the documentation no, but I had a lot of issues with stack too deep, and wanted to deploy it so the FE/BE could use it. I'll be fixing this, please dont resolve this comment so I wont forget it

contracts/chainlink/ccip/TLDCcipSender.sol
maikelordaz: Yes, the only one right now is USDC. To allow other ones must be also tokens listed in the CCIP, if not we must somehow swap it before building the CCIP message to other one supported by the protocol

contracts/chainlink/ccip/TLDCcipSender.sol
maikelordaz: Would not know what is better or easier, I can make it upgradeable, I might need maybe half work day, to do it as the CCIP contracts that are inherit must be rewritten in our source code. Build the calldata might be tricky, with the string. For example this is a calldata from it:

- `0xaa101130` First four bytes are the selector, easy one just keccak256 the function signature and take the first four bytes
- `0000000000000000000000000000000000000000000000000000000005f5e100` first 32-bytes word is the contribution
- `00000000000000000000000000000000000000000000000000000000000000a0` second 32-bytes word is a pointer to the offset in memory where the string starts, this one depends on a lot of things like modifiers, payable, solidity free memory pointer
- `0000000000000000000000000000000000000000000000000000000000000000` third 32-bytes word the parent address
- `00000000000000000000000087892a4c1c0bc10e32d34ffb8ef224bd600529a3` fourth 32-bytes word the member address
- `00000000000000000000000000000000000000000000000000000000047868c0` fifth the coupon amount
- `0000000000000000000000000000000000000000000000000000000000000013` sixth 32-bytes word the string length
- `546865204c6966652044414f20746573745f3200000000000000000000000000` seventh 32-bytes word the string content

Not exactly a fan of making this one upgradeable, but I see the importance of it and would say it can be the easiest, so I'll be doing it

contracts/chainlink/ccip/TLDCcipSender.sol
maikelordaz: Got it. Commit 5e1d2d7

contracts/referrals/ReferralGateway.sol
maikelordaz: Is this a name suggestion or a question? Well if it is a question, now `payContributionOnBehalfOf` can be called by the backend or the receiver contract. If it was a name suggestion, nice, I like it more than wrong caller, renamed it in commit b927065

contracts/chainlink/ccip/TLDCcipReceiver.sol
maikelordaz: Good point, I was also trying to get the user address from the calldata but it was tricky. I'll try it again so the user and owner can withdraw the tokens from a message Id, and also to allow the user to retry their messages by themselves, not only the owner

contracts/chainlink/ccip/TLDCcipSender.sol
maikelordaz: Commit a15308a

contracts/chainlink/ccip/TLDCcipReceiver.sol
maikelordaz: Commit 3522745

contracts/chainlink/ccip/TLDCcipReceiver.sol
maikelordaz: In this commit 336ec92 I changed a little bit the retry, now I get the user address from the data and the retry can be called with that address by either the owner or the user (let me know if you prefer to be only owner) made it with the address as I think is easier than storing somewhere each id, or check it against the user

maikelordaz: In this commit 1d3f7e6 I add a way for the user to recover the tokens, I'm assuming that it will be to the same address, even though we are in a different blockchain in this step, let me know if you prefer to have another input for an address to transfer the tokens

maikelordaz: In this commit da69049 I just refactor the previous two

contracts/chainlink/ccip/TLDCcipReceiver.sol
maikelordaz: Did not found any of it that made this, but I deleted in this commit d8ee564 because it was used only once, it did not make sense to have it there

contracts/chainlink/ccip/TLDCcipSender.sol
maikelordaz: I'm not able to avoid the stack too deep, the only way I can think right now is to make a decision on which token to use to pay the CCIP fees, the link or the source native blockchain token, that way I can remove the input `bool feesInLink` otherwise it have to be compiled via IR, it is not exactly a problem, it is just that the compilation that way is longer and the source code is big enough so it can take some time to compile on every change. Let me know what do you think

maikelordaz: Either way the benefit of not having it hardcoded is that there are some messages that fails that can be retried mannually in the CCCIP, this ones does not even get to the receiver contract. Some of those are because of gas, and you can increase the gas. What they propose with the gas limit as an input, is not to calculate it inside the contract, they propose to estimate it in a local test or deploy to testnet and see how much it is spend in the destination blockchain, or do it offchain with tenderly. They charge enough LINK or native token to pay that gas limit and dont reimburse anything. So I do agree it can not be hardcode it as it is

maikelordaz: This is a transaction in the receiver contract https://sepolia.arbiscan.io/tx/0x264880f44f4c94b433c10d6efd7fb8d9b471a8c485d3ca3efc4811e9f5a534a8 the gas spent was 876,893 close to the 1_000_000

contracts/chainlink/ccip/TLDCcipReceiver.sol
```diff
@@ -0,0 +1,293 @@
...
+    function _getUserMessageAndResolve(
+        address _user
+    ) internal returns (Client.Any2EVMMessage memory _message, bytes32 _messageId) {
+        _messageId = messageIdByUser[_user];
+
+        require(
+            failedMessages.get(_messageId) == uint256(ErrorCode.FAILED),
+            TLDCcipReceiver__MessageNotFailed(_messageId)
+        );
+        failedMessages.set(_messageId, uint256(ErrorCode.RESOLVED));
```
elallamTaka, line 289: Since these tokens are recovered how about we add a new error code RECOVERED to distinguish between these and complete payments? but this will require the status to be set outside this function

elallamTaka (changes requested):

contracts/chainlink/ccip/TLDCcipSender.sol
maikelordaz: Commits 283123e, 3041749 and bd62a76. FYI just removing the boolean was not enough, I removed one memory variable and solved it with that, but I was reading the solidity announcement channel, and I found since the last version they have improved the compilation time using the Intermidiate Representation pipeline, is not as fast as the legacy pipeline yet, but it works. I check it and compiling via ir takes 3 minutes, is an acceptable time in case we dont find a way to refactor something

contracts/chainlink/ccip/TLDCcipSender.sol
maikelordaz: Dont think so, as this is the sender contract and it will be in a different blockchain

elallamTaka (approved):

PR #59, opened 2025-02-06
feat: implement referral gateway into optimized architecture (TS-1270)

maikelordaz:


contracts/helpers/libraries/checks/AddressCheck.sol
elallamTaka, line 1: aren't checks technically errors? Or are you trying to compile common checks here across multiple contracts?

contracts/modules/manager/ModuleManager.sol
```diff
@@ -0,0 +1,135 @@
...
+
+    /**
+     * @notice Check if a given address is a module
+     */
+    function _checkIsModule(address _newModule) internal {
+        try ITLDModuleImplementation(_newModule).isTLDModule() returns (bytes4 funcSelector) {
+            if (funcSelector != ITLDModuleImplementation.isTLDModule.selector)
+                revert ModuleManager__NotModule();
+        } catch (bytes memory reason) {
+            if (reason.length == 0) revert ModuleManager__NotModule();
```
elallamTaka, line 132: What if the length of the reason != 0 ? If this function will return the same error in both cases then I don't think its necessary to check the length of the reason

contracts/modules/manager/ModuleManager.sol
```diff
@@ -0,0 +1,135 @@
...
+        // New module can not be address 0, can not be already a module, and the status must be disabled or enabled
+        require(newModule != address(0), ModuleManager__AddressZeroNotAllowed());
+        require(
+            addressToModule[newModule].moduleAddress == address(0),
+            ModuleManager__AlreadyModule()
+        );
+        require(
+            state == ModuleState.Disabled || state == ModuleState.Enabled,
+            ModuleManager__WrongInitialState()
+        );
```
elallamTaka, line 64: I know we talked about this during our call. Now that I think about it I cant think of a reason why we would add a module with the initial state of Disabled. I think to reduce the size of the function/ code we can assume that all newly added modules are enabled by default. This way we can remove this error as well. In the unusual case where we want to add a module that is disabled we can add it then change the state immediately.

contracts/modules/EntryModule.sol
```diff
@@ -91,71 +106,38 @@ contract EntryModule is
...
-
-        require(
-            newMember.memberState == MemberState.Inactive,
-            ModuleErrors.Module__WrongMemberState()
-        );
-        require(
-            contributionBeforeFee >= reserve.minimumThreshold &&
-                contributionBeforeFee <= reserve.maximumThreshold,
-            EntryModule__ContributionOutOfRange()
+        _onlyModuleState(ModuleState.Enabled);
```
elallamTaka, line 113: here I noticed that you added this check inside the external function, but for refund you added the _onlyModuleState(ModuleState.Enabled); inside the refund internal function. Is there a preference ?

elallamTaka (changes requested):
Hey Maikel, I was able to review the organization, libs, utils, module manager, entry module, I still have the member, prejoin, and revModule to review. My comments are mostly questions. I will do a second review tomorrow.

contracts/core/TakasureReserve.sol
```diff
@@ -36,17 +39,12 @@ contract TakasureReserve is
...
 
     uint256 private RPOOL; // todo: define this value
 
     bytes32 internal constant TAKADAO_OPERATOR = keccak256("TAKADAO_OPERATOR");
     bytes32 internal constant DAO_MULTISIG = keccak256("DAO_MULTISIG");
     bytes32 private constant PAUSE_GUARDIAN = keccak256("PAUSE_GUARDIAN");
-    bytes32 private constant MODULE_CONTRACT = keccak256("MODULE_CONTRACT");
-    bytes32 public constant REFERRAL_GATEWAY = keccak256("REFERRAL_GATEWAY");
 
     mapping(uint16 month => uint256 montCashFlow) public monthToCashFlow;
```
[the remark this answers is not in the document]
maikelordaz: Commit d072672

test/mocks/UsdcFaucet.sol
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * @title TLDCcipSender
+ * @author Maikel Ordaz
+ * @notice This contract will:
+ *          - Interact with the CCIP pprotocol
+ *          - Encode the data to send
+ *          - Send data to the Receiver contract
+ *          - Deployed in Avax (Mainnet and Fuji), Base (Mainnet and Sepolia), Ethereum (Mainnet and Sepolia),
+ *            Optimism (Mainnet and Sepolia), Polygon (Mainnet and Amoy)
```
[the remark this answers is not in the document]
maikelordaz: Wrong comments. Commit e811d52

contracts/helpers/libraries/checks/AddressCheck.sol
maikelordaz: My idea was to compile common checks in different contracts, when I wrote it the common one was that one, but now I can add the state check in each module, what do you say?

contracts/modules/manager/ModuleManager.sol
maikelordaz: true

maikelordaz: True. Commit 53f3b3a

contracts/modules/manager/ModuleManager.sol
maikelordaz: Fair. Commit 797e8d9

contracts/modules/EntryModule.sol
maikelordaz: Here is to revert before the external call to the reserve, to save gas, as require refund the gas that is left but not the gas already used

contracts/modules/manager/ModuleManager.sol
elallamTaka: We can also remove error ModuleManager__WrongInitialState(); if its not used anymore

contracts/helpers/libraries/checks/AddressCheck.sol
elallamTaka: question, whats best for size?

maikelordaz: If they everything is really common it wont make any difference. For example we put there checks for address zero, for module state and to check if the member is active, and we inherit the library in Module A, Module B and Module C. Module A and B use the three checks but module C uses only one

For Modules A and B there wont be any difference having them in the library or in the contract. But for Module C the size will be bigger

contracts/modules/manager/ModuleManager.sol
maikelordaz: Missed it. Commit c724209

elallamTaka (changes requested):

contracts/modules/PrejoinModule.sol
```diff
@@ -0,0 +1,854 @@
...
+
+        require(
+            _contribution >= MINIMUM_CONTRIBUTION && _contribution <= MAXIMUM_CONTRIBUTION,
+            PrejoinModule__ContributionOutOfRange()
+        );
+
+        if (_parent != address(0))
+            require(isMemberKYCed[_parent], PrejoinModule__ParentMustKYCFirst());
+
+        if (nameToDAOData[tDAOName].preJoinEnabled) {
```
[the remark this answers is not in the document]
maikelordaz: True, it is not usefull anymore. Will remove it

contracts/modules/PrejoinModule.sol
```diff
@@ -0,0 +1,854 @@
...
+        uint256 membershipDuration = 60 * 60 * 24 * 365 * 5; // 5 years
+
+        // Finally, we join the prepaidMember to the tDAO
+        IEntryModule(nameToDAOData[tDAOName].entryModule).joinPool(
+            newMember,
+            nameToDAOData[tDAOName].prepaidMembers[newMember].contributionBeforeFee,
+            membershipDuration
+        );
+
+        // TODO: Check if this is the correct amount to transfer
```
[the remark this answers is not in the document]
maikelordaz: Commit bf8c430

contracts/modules/PrejoinModule.sol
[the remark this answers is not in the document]
maikelordaz: Commit 63d503c

contracts/modules/PrejoinModule.sol
```diff
@@ -0,0 +1,854 @@
...
+     * @dev The member must have a tDAO assigned
+     */
+    function joinDAO(address newMember) external nonReentrant {
+        _onlyModuleState(ModuleState.Disabled);
+        // Initial checks
+        require(isMemberKYCed[newMember], PrejoinModule__NotKYCed());
+
+        require(
+            nameToDAOData[tDAOName].DAOAddress != address(0) &&
+                nameToDAOData[tDAOName].launchDate <= block.timestamp,
```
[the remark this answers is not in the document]
maikelordaz: Commit d9d059f

contracts/modules/EntryModule.sol
```diff
@@ -234,17 +214,115 @@ contract EntryModule is
...
+            _contributionBeforeFee: _newMember.contribution,
+            _contributionAfterFee: contributionAfterFee,
+            _memberWallet: _membersWallet,
+            _reserve: _reserve,
+            _takasureReserve: takasureReserve
+        });
+
+        _newMember.creditTokensBalance += mintedTokens;
+
+        emit TakasureEvents.OnMemberKycVerified(_newMember.memberId, _membersWallet);
```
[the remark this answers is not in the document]
maikelordaz: To track with the subgraph we should emit something about the KYC, I have in mind two alternatives:
1. Leave this event
2. Modify the `OnMemberCreated` event and add a boolean with the KYC status

As the code it is right now I think the second option looks better

contracts/core/TakasureReserve.sol
```diff
@@ -227,25 +214,21 @@ contract TakasureReserve is
...
         reserve.contributionToken = newContributionToken;
     }
 
-    function setNewFeeClaimAddress(
-        address newFeeClaimAddress
-    ) external onlyRole(TAKADAO_OPERATOR) notZeroAddress(newFeeClaimAddress) {
+    function setNewFeeClaimAddress(address newFeeClaimAddress) external onlyRole(TAKADAO_OPERATOR) {
+        AddressCheck._notZeroAddress(newFeeClaimAddress);
         feeClaimAddress = newFeeClaimAddress;
     }
```
elallamTaka, line 225: If the fee is sent to the operator multisig in the entry module, whats the point of having this here?

contracts/modules/EntryModule.sol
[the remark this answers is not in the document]
maikelordaz: Made this change in commit ef220e7

contracts/modules/EntryModule.sol
```diff
@@ -166,11 +148,13 @@ contract EntryModule is
...
-            ._getReserveAndMemberValuesHook(takasureReserve, memberWallet);
+    function setKYCStatus(address memberWallet) external onlyRole(ModuleConstants.KYC_PROVIDER) {
+        _onlyModuleState(ModuleState.Enabled);
+        AddressCheck._notZeroAddress(memberWallet);
+        (Reserve memory reserve, Member memory newMember) = _getReserveAndMemberValuesHook(
+            takasureReserve,
+            memberWallet
+        );
 
         require(!newMember.isKYCVerified, EntryModule__MemberAlreadyKYCed());
```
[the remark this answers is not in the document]
maikelordaz: True, commit 359765f

contracts/helpers/libraries/checks/AddressCheck.sol
maikelordaz: I added to this library the check for the module state in commit 7b11800

contracts/modules/PrejoinModule.sol
```diff
@@ -0,0 +1,854 @@
...
+        if (couponAmount > 0) emit OnCouponRedeemed(newMember, couponAmount);
+    }
+
+    /**
+     * @notice Set the KYC status of a member
+     * @param user The address of the member
+     * @dev Only the KYC_PROVIDER can set the KYC status
+     */
+    function setKYCStatus(address user) external onlyRole(KYC_PROVIDER) {
+        _onlyModuleState(ModuleState.Enabled);
```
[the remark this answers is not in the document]
maikelordaz: I made this with the states, but some commits ago and now I can not find the commit 🫥

contracts/core/TakasureReserve.sol
maikelordaz: Every like global variable is in the TakasureReserve, and readed if needed from the modules. Like the fee claim address and the Benefit Multiplier oracle address

elallamTaka (approved):

PR #60, opened 2025-02-20
revise: add user address as a param in ccip send message and fix amounts (TS-1363)

maikelordaz:
Short PR to add the user as an input when sending the message and to handle messages without transfers of tokens

### DIFF
--- a/contracts/helpers/chainlink/ccip/TLDCcipSender.sol
+++ b/contracts/helpers/chainlink/ccip/TLDCcipSender.sol
@@ -43,7 +43,12 @@ contract TLDCcipSender is Initializable, UUPSUpgradeable, Ownable2StepUpgradeabl
 
     event OnNewSupportedToken(address token);
     event OnBackendProviderSet(address backendProvider);
-    event OnTokensTransferred(bytes32 indexed messageId, uint256 indexed tokenAmount, uint256 fees);
+    event OnTokensTransferred(
+        bytes32 indexed messageId,
+        uint256 indexed tokenAmount,
+        uint256 indexed fees,
+        address user
+    );
 
     error TLDCcipSender__NotZeroTransfer();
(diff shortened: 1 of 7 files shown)


contracts/helpers/chainlink/ccip/TLDCcipSender.sol
```diff
@@ -43,8 +43,14 @@ contract TLDCcipSender is Initializable, UUPSUpgradeable, Ownable2StepUpgradeabl
...
-    event OnTokensTransferred(bytes32 indexed messageId, uint256 indexed tokenAmount, uint256 fees);
-
+    event OnTokensTransferred(
+        bytes32 indexed messageId,
+        uint256 indexed tokenAmount,
+        uint256 indexed fees,
+        address user
+    );
+
+    error TLDCcipSender__NotZeroTransfer();
```
elallamTaka, line 53: Not zero transfer kinda sounds like it needs to be a zero transfer for the function to work. maybe only `ZeroTransferNotAllowed `?

contracts/helpers/chainlink/ccip/TLDCcipSender.sol
```diff
@@ -148,8 +154,10 @@ contract TLDCcipSender is Initializable, UUPSUpgradeable, Ownable2StepUpgradeabl
         uint256 contribution,
         string calldata tDAOName,
         address parent,
+        address newMember,
         uint256 couponAmount
     ) external returns (bytes32 messageId) {
+        require(amountToTransfer > 0, TLDCcipSender__NotZeroTransfer());
```
elallamTaka, line 160: Should we also check that the amountToTransfer <= contribution? I cant imagine a case where the user transfers an amount > than the contribution as of now

elallamTaka (approved):

PR #70, opened 2025-04-14
chore: sc rename set kyc status function (TS-1583)

maikelordaz:
Rename function `setKYCStatus` to `approveKYC`

spiderman-meme-1630319444


elallamTaka (commented):
That made me laugh hahaha. Before you deploy this we will need to update the call in the backend as well. Can you please add this as part of the migration doc, thanks

elallamTaka (approved):

PR #71, opened 2025-04-14
revise: validate launch date in update launch date (TS-1590)

maikelordaz:
Validate `launchDate` is previous to current one when updating


elallamTaka (approved):
The thing about this change is that it wont let us "start earlier" if the launch date is changed to a closer date. But tbh its not likely to happen for this launch, but must be revised for future DAOs architecture.

PR #76, opened 2025-04-14
chore: remove dao name instances (TS-1597)

maikelordaz:
Not only removed the DAOName from inputs, also made it constant, it is doable because it was not deployed yet so it was not taking any storage slot. 
The only thing with that is that inn uat we'll be having a new DAO, but as we are testing features dont think it is a problem


elallamTaka:
This PR requires the backend and frontend attention since the function call removed the DAO name. Thus we can leave it for last

elallamTaka (approved):
Please add a note in the migration doc so that the FE and BE can remove the input from their calls.

PR #79, opened 2025-04-15
revise: no kyc to refund users (TS-1601)

maikelordaz:


contracts/modules/EntryModule.sol
```diff
@@ -233,7 +233,7 @@ contract EntryModule is
         );
 
         require(!newMember.isKYCVerified, EntryModule__MemberAlreadyKYCed());
-        require(newMember.contribution > 0, EntryModule__NoContribution());
+        require(newMember.contribution > 0 && !newMember.isRefunded , EntryModule__NoContribution());
```
elallamTaka, line 236: To confirm, we decided to create a new member object each time someone joins, rather than setting isRefunded to false. Thus isRefunded should not be reset after a refund.

elallamTaka (approved):

elallamTaka (approved):
The change approved but it would be best to create an error message specifically for the refunded check. This is to enhance comms with the user.

PR #80, opened 2025-04-15
revise: fix duplicates from prejoiners (TS-1602)

maikelordaz:


contracts/modules/EntryModule.sol
```diff
@@ -151,6 +150,7 @@ contract EntryModule is
             takasureReserve,
             membersWallet
         );
+        if (!newMember.isRefunded) require(newMember.wallet == address(0), EntryModule__AlreadyJoined());
```
elallamTaka, line 153: This checks needs to be added to `joinPoolOnBehalfOf `as well in case backend tries to call it twice.

elallamTaka (changes requested):

contracts/modules/EntryModule.sol
maikelordaz: Done. Commit 7ba6b71

elallamTaka (approved):

PR #84, opened 2025-04-16
revise: restrict refund call (TS-1624)

maikelordaz:
For consistency

### DIFF
--- a/contracts/modules/EntryModule.sol
+++ b/contracts/modules/EntryModule.sol
@@ -517,7 +517,7 @@ contract EntryModule is
             takasureReserve,
             _memberWallet
         );
-        require(_memberWallet == _member.wallet || hasRole(ModuleConstants.TAKADAO_OPERATOR, msg.sender), EntryModule__NotAuthorizedCaller());
+        require(_memberWallet == msg.sender || hasRole(ROUTER, msg.sender) || hasRole(ModuleConstants.TAKADAO_OPERATOR, msg.sender), EntryModule__NotAuthorizedCaller());
 
         // The member should not be KYCed neither already refunded
         require(!_member.isKYCVerified, EntryModule__MemberAlreadyKYCed());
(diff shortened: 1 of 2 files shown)


contracts/modules/EntryModule.sol
```diff
@@ -517,6 +517,7 @@ contract EntryModule is
             takasureReserve,
             _memberWallet
         );
+        require(_member.wallet == _memberWallet || hasRole(ModuleConstants.TAKADAO_OPERATOR, msg.sender), EntryModule__NotAuthorizedCaller());
```
elallamTaka, line 520: I think you meant `_memberWallet == msg.sender`. Also, the constant has changed based on this https://github.com/TakafulDAO/takasure.sc/pull/73

elallamTaka (changes requested):

contracts/modules/EntryModule.sol
maikelordaz: Commit ac0f590

test/unit/03-modules/02-entry-module/03_Refund_EntryModuleTest.t.sol
```diff
@@ -165,22 +165,4 @@ contract Refund_EntryModuleTest is StdCheats, Test, SimulateDonResponse {
         assertEq(aliceAfterFirstJoinBeforeRefund.memberId, aliceAfterRefund.memberId);
         assertEq(aliceAfterRefund.memberId, aliceAfterSecondJoin.memberId);
     }
-
-    function testEntryModule_refundCalledByAnyone() public {
```
elallamTaka, line 169: Rather than removing this test altogether, I suggest you convert it to  `testEntryModule_refundRevertIfCalledByWrongSender()`
To test that no user gets a refund unless he is the address calling the function.

elallamTaka (approved):

PR #85, opened 2025-04-21
revise: refund math (TS-1625)

maikelordaz:
If someone used coupons and ask for a refund it will refund the coupon pool. The next transfer needs to be handled offchain


elallamTaka (approved):
Before merging this PR please add a comment over the refund function that explains the behavior, this would be useful for future audits example: 
Note that users who used a full or partial coupon will result in the full refund transferring to the coupon pool. Customer support is then required to refund users who used partial coupons. This behavior is intentional.

PR #88, opened 2025-04-28
feat: upload pioneer nft data to ipfs (TS-1532)

maikelordaz:
## Important

Please review [PR 64](https://github.com/TakafulDAO/takasure.sc/pull/64) first

## What is implemented here

- Deploy scripts, although with other PRs will be needed to reorganize the ones written here
- Set the base uri
- Change the way of minting from PR 64

## What changed in the minting process

Now the NFTs start at token id 00000, in the previous PR it was on id 1. This change was needed after I realized how the images where named

## Regarding the URI

I upload the imaged in my local node the CID is QmYLyTRp3uUN8ryGw2NaLPoudicgSJDr4E5DGTn8tLj8gP and the base URI is with the normal `https://ipfs.io/ipfs/` so you can view the images at `https://ipfs.io/ipfs/QmYLyTRp3uUN8ryGw2NaLPoudicgSJDr4E5DGTn8tLj8gP/` this CID just points to the images, and I used it to make some tests, but to display the images correctly for example in Metamask or OpenSea it is needed the metadata with the proper standard


MohammedTaka (changes requested):
The smart contract should not point directly at the images, rather the jsons.

The jsons start correctly with 0, 1, 2, 3 etc. so there is no need to accomdate for the fact that images start with 00000, 00001 etc.

maikelordaz:
Reference comment from canceled PR https://github.com/TakafulDAO/takasure.sc/pull/64#discussion_r2154401754
Fixed in commit d2b0b2267e9f3a01dd7972229b65cdb1b85bdac6

PR #93, opened 2025-05-01
revise: fix approval (TS-1697)

maikelordaz:


contracts/helpers/chainlink/ccip/TLDCcipReceiver.sol
```diff
@@ -122,11 +121,9 @@ contract TLDCcipReceiver is CCIPReceiver, Ownable2Step {
 
         // Get the old protocol gateway address and set the approval to 0
         address oldProtocolGateway = protocolGateway;
-        usdc.approve(oldProtocolGateway, 0);
 
         // Set the new protocol gateway address and approve the maximum amount
```
elallamTaka, line 125: This comment is outdated now

contracts/helpers/chainlink/ccip/TLDCcipReceiver.sol
```diff
@@ -122,11 +121,9 @@ contract TLDCcipReceiver is CCIPReceiver, Ownable2Step {
 
         // Get the old protocol gateway address and set the approval to 0
         address oldProtocolGateway = protocolGateway;
-        usdc.approve(oldProtocolGateway, 0);
```
elallamTaka, line 125: Setting the approval to 0 here is still valid. Say a users payment failed. This it is stuck here. If we change the gateway we want to retry from the new one, thus we must remove any hanging approvals to the old gateway.

elallamTaka (changes requested):

contracts/helpers/chainlink/ccip/TLDCcipReceiver.sol
maikelordaz: Commit 8774097

contracts/helpers/chainlink/ccip/TLDCcipReceiver.sol
maikelordaz: 🤔 I don't get it

maikelordaz: Forget it, I get it now

maikelordaz: But the approval is from this contract allowing the `protocolGateway` to use the tokens, not from the user. The user approves the `TLDCcipSender` to use the tokens, this tokens are bridged and received (and owned) by the `TLDCcipReceiver` and the `TLDCcipReceiver` is setting the approval inside `_ccipReceive` if the transaction fails the approval will also fail. And if this happens and the user wants to try again, the approval happens again in `retryFailedMessage`

elallamTaka (approved):

PR #97, opened 2025-05-02
revise token should be non transferable (TS-1715)

maikelordaz:


maikelordaz:
Closing as we are no using the token as credits anymore

PR #102, opened 2025-05-04
revise: sc validate amount to bridge (TS-1716)

maikelordaz:


test/unit/05-chainlink/00-ccip/TLDCcipTest.t.sol
```diff
@@ -375,7 +375,7 @@ contract TLDCcipTest is Test {
     }
 
     function testSenderSendMesageRevertsIfThereIsCouponAndCallerIsNotBackend() public setToken {
-        uint256 amountToTransfer = 100e6;
+        uint256 amountToTransfer = 50e6;
```
elallamTaka, line 378: Whats the benefit behind changing the values for this test?

maikelordaz: That test is checking the caller. And as the amount validation is before the caller validation, then the amount validation has to pass.

elallamTaka (approved):

PR #105, opened 2025-05-04
revise: sc fix burns from ts token (TS-1720)

maikelordaz:


maikelordaz:
Closing as we are no using the token as credits anymore

PR #114, opened 2025-05-06
revise: fix coupon transfers (TS-1735)

maikelordaz:


contracts/modules/EntryModule.sol
elallamTaka, line 1: Hey Maikel, I dont think the _amountToTransferFromMember is properly calculated. For example, say the user wants to pay 250 USDC and he has a 50 USDC coupon. Then according to the reqs, we must withdraw from his wallet (imagine reserve fee is 20%):
250 - 50 (coupon) - 20% of 200 (because we must apply discounts after coupon deduction) = 160 USDC. 

So we must withdraw 160 USDC from the user . But in the code the calculation is:
200 (contributionAfterFee) - 10 (discount) - 50 (coupon) = 140 USDC. 

Can you please write a test for this to confirm. The test must call payOnBehalf with 250 and a coupon of 50.

maikelordaz: I read this comment and I thought it was on the PR from the ReferralGateway and think I made it there. Let me check it here also

maikelordaz: With this test I realizaed there was an issue declaring the default values. Fixed in this commit 4a26d055a27f1bdd59b6c96d38b2fc7905a153f4

maikelordaz: And made the test in this commit 4cf2bccb2d10c1e92eda781069383b046e493cd8

elallamTaka (approved):

PR #115, opened 2025-05-06
revise: avoid recalculations (TS-1736)

maikelordaz:


contracts/modules/EntryModule.sol
```diff
@@ -631,21 +628,32 @@ contract EntryModule is
...
+        if (_isKYCVerified && !_isRefunded) {
+            // Then the call comes from approveKYC, no need to update the claimAddAmount, contribution and fee
+            claimAddAmount = _member.claimAddAmount;
+            contribution = _member.contribution;
+            fee = _member.totalServiceFee;
         } else {
-            userMembershipDuration = ModuleConstants.DEFAULT_MEMBERSHIP_DURATION;
+            // Otherwise it is a refunded member that is rejoining the pool
+            claimAddAmount = ((normalizedContributionBeforeFee - feeAmount) * (100 - _drr)) / 100;
+            contribution = normalizedContributionBeforeFee;
```
elallamTaka, line 647: Same thing here, `_calculateAmountAndFees` is not called so these are 0 ?

elallamTaka (changes requested):
I dont see how this PR addressess the issue mentioned in the Hacken issue F-2025-9835. The issue says that the recorded contribution inside of approveKYC will be inconsistent if the operator changes the fees before the KYC was approved. Thus the fix is to store the fee that was calculated for the user when he joined. OR to call transfer service fee only after the user does KYC. Moving this outside of `_transferContributionToModule`. 
I think we need to move it anyways because the name of the function refers to the transfer of the contribution, not the fees. 
```
            // Transfer the service fee to the fee claim address
            contributionToken.safeTransferFrom(
                _memberWallet,
                takasureReserve.feeClaimAddress(),
                feeAmount
            );
```

Once we move it outside then calling `_calculateAmountAndFees `isnt a problem anymore.  This also means that by design, its fine if the op changes fees suddenly. The applied fee would always be the new one for users who approve their KYC after the change.

maikelordaz:
The use of the transients variables here was mostly a way to avoid the stack too deep and to reduce the contract size at the same time, but I'm thinking now with the upcoming changes this contract might reduce in size and the I can change the internal function to receive the inputs as structs, this could help to reduce the complexity. But will be a work for the new PRs

PR #118, opened 2025-05-06
chore: deprecate objective amount (TS-1741)

maikelordaz:


contracts/modules/PrejoinModule.sol
```diff
@@ -192,10 +190,10 @@ contract PrejoinModule is
         nameToDAOData[DAOName].referralDiscount = isReferralDiscountEnabled;
         nameToDAOData[DAOName].DAOAdmin = operator;
         nameToDAOData[DAOName].launchDate = launchDate;
-        nameToDAOData[DAOName].objectiveAmount = objectiveAmount;
+        nameToDAOData[DAOName].objectiveAmount = 0;
```
elallamTaka, line 193: Is this necessary? I think its set to 0 by default

maikelordaz: True, changed in commit bde0b29df62e968dc9dbf8db45811524c13db82d

elallamTaka (approved):

PR #123, opened 2025-05-09
revise: try/catch for parent rewards (TS-1733)

maikelordaz:


contracts/modules/PrejoinModule.sol
```diff
@@ -379,9 +385,12 @@ contract PrejoinModule is
...
 
-            usdc.safeTransfer(parent, parentReward);
-
-            emit OnParentRewarded(parent, layer, user, parentReward);
+            try usdc.transfer(parent, parentReward) {
+                // Emit the event only if the transfer was successful
+                emit OnParentRewarded(parent, layer, user, parentReward);
+            } catch {
+                emit OnParentTransferFailed(parent, layer, user, parentReward);
+            }
```
elallamTaka, line 393: Here after failing we must also recredit the parent with the reward.
`nameToDAOData[tDAOName].prepaidMembers[parent].parentRewardsByChild[user] = parentReward ;`
Also, we must add an Op only function that transfers the reward of a parent to an address specified as input. This way we can get the money out of the contract. Technically we must also have a fallback to "transfer" a membership to another address if this happens, but I still think that the risk is too small to handle as of now.

elallamTaka (changes requested):

contracts/modules/PrejoinModule.sol
maikelordaz: Commit b0db2d35a57a427c16a4c7738788cf6cb4231ae0, also wrote it in EntryModule in commit 4fdbdb7fde6382b54171964ef62fcdb1ef616aeb

elallamTaka (approved):

PR #124, opened 2025-05-22
revise: change in referral gateway (TS-1812)

maikelordaz:


elallamTaka (changes requested):

contracts/referrals/ReferralGateway.sol
```diff
@@ -650,135 +608,126 @@ contract ReferralGateway is
...
-        uint256 _couponAmount
+        uint256 _couponAmount,
+        bool _isDonated
     ) internal whenNotPaused nonReentrant returns (uint256 _finalFee, uint256 _discount) {
         uint256 realContribution;
 
-        if (_couponAmount > _contribution) realContribution = _couponAmount;
-        else realContribution = _contribution;
+        if (_isDonated) {
+            realContribution = MINIMUM_CONTRIBUTION;
```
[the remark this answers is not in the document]
maikelordaz: Commit 007603adda361cc0b3b080d562012622922e7168

contracts/referrals/ReferralGateway.sol
```diff
@@ -364,41 +325,46 @@ contract ReferralGateway is
...
-        uint256 couponAmount
+        uint256 couponAmount,
+        bool isDonated
     ) external returns (uint256 finalFee, uint256 discount) {
         _onlyCouponRedeemerOrCcipReceiver();
+        if (isDonated)
+            require(contribution == MINIMUM_CONTRIBUTION, ReferralGateway__InvalidContribution());
+
+        if (couponAmount > 0)
+            require(couponAmount <= contribution, ReferralGateway__InvalidContribution());
```
elallamTaka, line 338: I dont think an if is require here, using the require immediately is still correct

maikelordaz: Commit d2ab4afe7e82d588f2d3db419638488fa059f52f

contracts/referrals/ReferralGateway.sol
```diff
@@ -650,150 +612,136 @@ contract ReferralGateway is
...
+
         if (_parent != address(0))
             require(isMemberKYCed[_parent], ReferralGateway__ParentMustKYCFirst());
 
-        if (nameToDAOData[_tDAOName].preJoinEnabled) {
+        if (nameToDAOData[daoName].preJoinEnabled) {
             // We check if the member already exists
             require(
-                nameToDAOData[_tDAOName].prepaidMembers[_newMember].contributionBeforeFee == 0,
+                nameToDAOData[daoName].prepaidMembers[_newMember].contributionBeforeFee == 0,
```
elallamTaka, line 744: Why is this condition only checked inside of this if statement? Once we disable prejoin discount this should still hold true

elallamTaka (changes requested):

contracts/referrals/ReferralGateway.sol
maikelordaz: True. Commit d3651f2428465e4bdf40b86c6627ce2173218ae1

elallamTaka: Can you run the common tests with prejoin on and prejoin off. A test that tries to have a user double join after prejoin is disabled would have caught this. Otherwise we will have double joins when we disable prejoin

PR #125, opened 2025-05-29
revise: parent reward changes (TS-1770)

maikelordaz:
## Changes in modules

1. Removed `EntryModule`
2. Create `SubscriptionModule`
2.1 It will handle the subscriptions
2.2 It will handle the refunds
2.3 It will hold the payment until the user is KYCed
2.4 Only user related actions (although ones are called by us)
2.5 Renamed `joinPool` to `paySubscription`
3. Create `KYCModule`
3.1 It will KYC a user
3.2 It will run the algorithms
3.3 It will trigger the contribution transfer to the reserve
3.4 Only admin functions

## Changes to previous flow

1. Now refund deleted the user except for the id
2. This approach reduced calls like `_updateMember` is not necessary anymore

## Tests

The tests actually check the same thing, but the set up is different because of the different modules (this is what took more time)

## To Do

For other PRs is still pending this:

- There should be an extra module for the Life Benefit
- The algorithms should only run if someone applies to that benefit
- The Life Benefit Module should be able to be called after paying the subscription
- The Life Benefit should also be accesible from the `paySubscription` function been thinking on having an Enum in the inputs (Life and Burial benefits) if any benefit is selected since initial payment then the function allows to pay more than 25 and the algorithms runs on KYC, if non benefit is selected then the payment is restricted to 25USDC and non algorithms are performed

## Notes

Prior this changes `TakasureReserve.sol` and `EntryModule.sol` where off the size limit, with this it is back on size, and with the implementition of the `AddressManager` or `AccessControlManager` (open to suggestions) should decrease even more


contracts/modules/KYCModule.sol
```diff
@@ -0,0 +1,243 @@
+//SPDX-License-Identifier: GPL-3.0
+
+/**
+ * @title EntryModule
+ * @author Maikel Ordaz
+ * @notice This contract manage all the process to become a member
+ * @dev It will interact with the TakasureReserve contract to update the values
+ * @dev Upgradeable contract with UUPS pattern
```
elallamTaka, line 8: the natspec here is incorrect

contracts/modules/KYCModule.sol
```diff
@@ -0,0 +1,243 @@
...
+        AddressAndStates._onlyModuleState(moduleState, ModuleState.Enabled);
+        AddressAndStates._notZeroAddress(memberWallet);
+
+        (Reserve memory reserve, Member memory newMember) = _getReserveAndMemberValuesHook(
+            takasureReserve,
+            memberWallet
+        );
+
+        require(!newMember.isKYCVerified, KYCModule__MemberAlreadyKYCed());
+        require(newMember.contribution > 0 && !newMember.isRefunded, KYCModule__NoContribution());
```
elallamTaka, line 110: I suggest to change `KYCModule__NoContribution` to `KYCModule__ContributionRequired()`

contracts/modules/KYCModule.sol
```diff
@@ -0,0 +1,243 @@
...
+
+    function _calculateAmountAndFees(uint256 _contributionBeforeFee, uint256 _fee) internal {
+        // The minimum we can receive is 0,01 USDC, here we round it. This to prevent rounding errors
+        // i.e. contributionAmount = (25.123456 / 1e4) * 1e4 = 25.12USDC
+        normalizedContributionBeforeFee =
+            (_contributionBeforeFee / ModuleConstants.DECIMAL_REQUIREMENT_PRECISION_USDC) *
+            ModuleConstants.DECIMAL_REQUIREMENT_PRECISION_USDC;
+        feeAmount = (normalizedContributionBeforeFee * _fee) / 100;
+        contributionAfterFee = normalizedContributionBeforeFee - feeAmount;
+    }
```
elallamTaka, line 180: This may be required when users do a recurring payment, we always deduct the service fee for each yearly payment. Perhaps create a fees helper lib?

contracts/modules/KYCModule.sol
```diff
@@ -0,0 +1,243 @@
...
+            _benefitMultiplier,
+            normalizedContributionBeforeFee,
+            feeAmount,
+            userMembershipDuration,
+            block.timestamp
+        );
+        return _member;
+    }
+
+    function _getBenefitMultiplierFromOracle(
```
elallamTaka, line 222: How about we get rid of the BM Oracle all together in this PR and have a setter that is called by the backend to set a users BM to reduce the complexity of this version. (otherwise we can do it in another PR)

contracts/modules/SubscriptionModule.sol
```diff
@@ -0,0 +1,621 @@
...
+    ) external initializer {
+        __UUPSUpgradeable_init();
+        __AccessControl_init();
+        __ReentrancyGuardTransient_init();
+
+        takasureReserve = ITakasureReserve(_takasureReserveAddress);
+        bmConsumer = IBenefitMultiplierConsumer(takasureReserve.bmConsumer());
+
+        address takadaoOperator = takasureReserve.takadaoOperator();
+        address moduleManager = takasureReserve.moduleManager();
```
elallamTaka: These are not used anywhere in this contract

contracts/modules/SubscriptionModule.sol
```diff
@@ -0,0 +1,621 @@
...
+        address parentWallet,
+        uint256 contributionBeforeFee,
+        uint256 membershipDuration
+    ) external nonReentrant {
+        AddressAndStates._onlyModuleState(moduleState, ModuleState.Enabled);
+
+        require(
+            msg.sender == referralGateway ||
+                hasRole(ModuleConstants.ROUTER, msg.sender) ||
+                msg.sender == memberWallet,
```
elallamTaka, line 144: Maybe its the formatter you are using but the indentation here can be prettier if the other lines align with the first requirement (like in the attachment)  
image

contracts/modules/SubscriptionModule.sol
```diff
@@ -84,16 +81,17 @@ contract EntryModule is
 
         takasureReserve = ITakasureReserve(_takasureReserveAddress);
         bmConsumer = IBenefitMultiplierConsumer(takasureReserve.bmConsumer());
+
         address takadaoOperator = takasureReserve.takadaoOperator();
         address moduleManager = takasureReserve.moduleManager();
```
elallamTaka, line 86: These are not used anywhere in this contract

elallamTaka (changes requested):

contracts/modules/SubscriptionModule.sol
```diff
@@ -463,84 +399,70 @@ contract EntryModule is
...
             takasureReserve,
             _memberWallet
         );
-        require(_memberWallet == msg.sender || hasRole(ModuleConstants.ROUTER, msg.sender) || hasRole(ModuleConstants.OPERATOR, msg.sender), ModuleErrors.Module__NotAuthorizedCaller());
 
+        require(_memberWallet == msg.sender || hasRole(ModuleConstants.ROUTER, msg.sender) || hasRole(ModuleConstants.OPERATOR, msg.sender), ModuleErrors.Module__NotAuthorizedCaller());
         // The member should not be KYCed neither already refunded
-        require(!_member.isKYCVerified, EntryModule__MemberAlreadyKYCed());
-        require(!_member.isRefunded, EntryModule__NothingToRefund());
+        require(!_member.isKYCVerified, SubscriptionModule__MemberAlreadyKYCed());
```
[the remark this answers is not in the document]
maikelordaz: Yes, read about it, commit in 8b5c5e9ae3b6f53e0f50086af7613df55cce04ca but this has to be revisited maybe in the PR when working on the algorithmic and model changes, because also read that if the user applies for the Life protection it can not be refunded. Also thinking that now with new requirements this should be an admin function because also read that can not refund if applies for the cripto card, and that would be something outside the SC scope

contracts/modules/SubscriptionModule.sol
maikelordaz: The `takadaoOperator` is used only inside the initialize function to assign roles, it is not a global var, but I changed the declaration of the `moduleManager` because it is used only one in this commit 881f960fb6765b28e3f3eb969237d07f4e37f7f6

contracts/modules/SubscriptionModule.sol
maikelordaz: I have to change the formatter I'm using some functions declarations have similar issues

contracts/modules/KYCModule.sol
maikelordaz: Change it here `f1ef212b4af9384d7d69fafcb014d4fce4af0a4b`

contracts/modules/KYCModule.sol
maikelordaz: Commit 953da6b30db6a48183e74ed1d603f215915e688f

contracts/modules/KYCModule.sol
maikelordaz: I have like opposite opinions on this. On one side I agree with it because it is definitely complex, and the fact that needs two calls makes it worst, but on the other side I think it will add more centralization if we remove it, because if the BM is given by us then it will restrict the caller, the only I could thought about it was:

1. User pay subscription and join life benefit, we give a place holder for the BM here, it does not matter if it is called `paySubscription` or `paySubscriptionOBehalfOf`
2. Backend KYC the user and here is where we set the real BM, as an input, also here we can have an extra input so the backend tell the smart contract the BMA method to use (maybe this one optional if there is no change), so it would be something like `approveKYC(address memberWallet, uint256 benefitMultiplier, BMAMethod bmaMethod)` where BMAMethod it is an enum

What do you think?

PS: I'm still more on the side of remove it

contracts/modules/SubscriptionModule.sol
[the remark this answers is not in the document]
elallamTaka: Yup you are correct, we can revisit refund in another PR, once we apply the address manager we will go over the entire code base and trace it back to each req.

contracts/modules/KYCModule.sol
```diff
@@ -0,0 +1,209 @@
...
+        address _takasureReserve,
+        uint256 _contributionAfterFee
+    ) internal override {
+        subscriptionModule.transferContributionAfterKyc(
+            _contributionToken,
+            _memberWallet,
+            _takasureReserve,
+            _contributionAfterFee
+        );
+    }
```
elallamTaka, line 203: This function s not being called anywhere in the KYC module, how do you intend to use it?

contracts/modules/SubscriptionModule.sol
```diff
@@ -624,50 +554,6 @@ contract EntryModule is
...
-            normalizedContributionBeforeFee,
-            feeAmount,
-            userMembershipDuration,
-            block.timestamp
-        );
-
-        return _member;
-    }
-
     function _transferContribution(
```
elallamTaka, line 557: I suggest we rename this as _transferContributionToReserve

contracts/modules/SubscriptionModule.sol
```diff
@@ -209,96 +209,31 @@ contract EntryModule is
...
-            takasureReserve,
-            memberWallet
+    function transferContributionAfterKyc(
+        IERC20 contributionToken,
+        address memberWallet,
+        address takasureReserveAddress,
+        uint256 contributionAfterFeeAmount
+    ) external {
+        require(
+            moduleManager.isActiveModule(msg.sender),
```
elallamTaka, line 227: This isnt safe, anyone can create a module that follows the modules implementation and set it as active to call this function. Rather, it should check specifically that the caller is the KYC module, which should be fetched from the AddressManager

contracts/modules/SubscriptionModule.sol
```diff
@@ -132,26 +134,28 @@ contract EntryModule is
...
+            msg.sender == referralGateway ||
+                hasRole(ModuleConstants.ROUTER, msg.sender) ||
+                msg.sender == memberWallet,
+            ModuleErrors.Module__NotAuthorizedCaller()
+        );
 
-        if (msg.sender == prejoinModule) {
-            _joinFromPrejoinModule(
+        if (msg.sender == referralGateway) {
+            _joinFromReferralGateway(
```
elallamTaka, line 158: I highly recommend we remove `_joinFromReferralGateway` from `paySubscription` and make it an external function by itself. This separates the concern of users who joined from the gateway and dedicates `paySubscription` to new users joining the association only. The code also becomes easier to review, and once the gateway is deprecated, its only a matter of removing the external function, rather than editing `paySubscription`.

contracts/modules/SubscriptionModule.sol
```diff
@@ -177,28 +181,24 @@ contract EntryModule is
...
 
-        uint256 benefitMultiplier = _getBenefitMultiplierFromOracle(memberWallet);
+        // Check if the coupon amount is valid
+        require(couponAmount <= contributionBeforeFee, SubscriptionModule__InvalidContribution());
 
-        _calculateAmountAndFees(contributionBeforeFee, reserve.serviceFee);
+        // Check caller
+        _onlyCouponRedeemerOrCcipReceiver();
 
         _join(
```
elallamTaka, line 203: _join makes checks and stores values that are no longer relevant. For example, it checks the min and mac contribution based on the reserve which is not relevant anymore since the subscription cost is 25 only. It also stores the DRR of the reserve and BM which is also not relevant as those only apply when the money actually gets deposited into the reserve. Its only deposited in these 2 conditions:
* user applied for ALP
* user does not apply for ALP within 30 days, then their money goes in as a donation

contracts/modules/KYCModule.sol
maikelordaz: What do you think about that approach? I can remove it completely on the PR to apply the model changes

contracts/modules/SubscriptionModule.sol
maikelordaz: Yes, but I was thinking to be apply those kind of changes, like new requirements, model changes, algorithmic changes and fix the tests after it in the ticket https://takadao.atlassian.net/browse/TS-1773

contracts/modules/SubscriptionModule.sol
```diff
@@ -319,11 +254,32 @@ contract EntryModule is
...
+        if (!newMember_.isRefunded) {
+            require(newMember_.wallet == address(0), SubscriptionModule__AlreadyJoined());
+        } else {
+            require(
+                newMember_.memberId != 0 && newMember_.wallet == _memberWallet,
+                ModuleErrors.Module__WrongMemberState()
+            );
+        }
+
+        benefitMultiplier_ = _getBenefitMultiplierFromOracle(_memberWallet);
```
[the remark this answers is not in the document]
maikelordaz: Yes, but I was thinking to be apply those kind of changes, like new requirements, model changes, algorithmic changes and fix the tests after it in the ticket https://takadao.atlassian.net/browse/TS-1773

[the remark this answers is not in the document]
maikelordaz: Either way each day I'm thinking more and more that we definitely need to remove the oracle

contracts/modules/SubscriptionModule.sol
maikelordaz: Commit 2cc9fbe4eca005b7774c6c4dd255a60f27291206

contracts/modules/KYCModule.sol
maikelordaz: I reference this comment in ticket 1773

contracts/modules/SubscriptionModule.sol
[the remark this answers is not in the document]
maikelordaz: I reference this comment in ticket 1773

contracts/modules/SubscriptionModule.sol
maikelordaz: I reference this comment in ticket 1773

elallamTaka (approved):

PR #127, opened 2025-06-19
revise: nft contract (TS-1925)

maikelordaz:


contracts/tokens/RevShareNFT.sol
```diff
@@ -0,0 +1,281 @@
...
+
+    uint256 public totalSupply; // Total supply of NFTs minted
+    uint256 private mintedToOperator; // Updated on transfers from the operator
+    uint256 private mintedToPioneers; // Updated on mints
+
+    mapping(address pioneer => uint256 timestamp) public pioneerMintedAt;
+
+    uint256 public constant MAX_SUPPLY = 18_000;
+    uint256 private constant OPERATOR_RESERVED_TOKENS = 9_180; // 51% of the total supply reserved for operator
+    uint256 private constant PIONEERS_RESERVED_TOKENS = 8_820; // 49% of the total supply reserved for pioneers
```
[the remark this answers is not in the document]
maikelordaz: Commit ec2c574

contracts/tokens/RevShareNFT.sol
```diff
@@ -0,0 +1,281 @@
...
+    using SafeERC20 for IERC20;
+
+    ITakasureReserve private takasureReserve;
+
+    address public operator;
+    string public baseURI; // Base URI for the NFTs
+
+    uint256 public totalSupply; // Total supply of NFTs minted
+    uint256 private mintedToOperator; // Updated on transfers from the operator
+    uint256 private mintedToPioneers; // Updated on mints
```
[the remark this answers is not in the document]
maikelordaz: Commit ec2c574

contracts/tokens/RevShareNFT.sol
```diff
@@ -0,0 +1,281 @@
...
+
+    mapping(address pioneer => uint256 timestamp) public pioneerMintedAt;
```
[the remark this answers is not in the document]
maikelordaz: I do agree it is missing the id but only to be used from now until the revenue contract is deployed. After that this variable is not needed anymore as calculating revenues that way on chain is not the best approach

contracts/tokens/RevShareNFT.sol
```diff
@@ -0,0 +1,281 @@
...
+    UUPSUpgradeable,
+    Ownable2StepUpgradeable,
+    ReentrancyGuardTransientUpgradeable,
+    ERC721Upgradeable
+{
+    using SafeERC20 for IERC20;
+
+    ITakasureReserve private takasureReserve;
+
+    address public operator;
```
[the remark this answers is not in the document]
maikelordaz: Commit 880e683

contracts/tokens/RevShareNFT.sol
```diff
@@ -0,0 +1,281 @@
...
+        baseURI = _baseURI;
+
+        totalSupply = OPERATOR_RESERVED_TOKENS - 1; // We assume that the operator has 9180 tokens already minted and the tokenId starts at 0
+    }
+
+    /*//////////////////////////////////////////////////////////////
+                                SETTINGS
+    //////////////////////////////////////////////////////////////*/
+
+    function setTakasureReserve(address _takasureReserve) external onlyOwner {
```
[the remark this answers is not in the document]
maikelordaz: Commit 515304b

contracts/tokens/RevShareNFT.sol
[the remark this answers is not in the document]
maikelordaz: Commit 6c36f85

contracts/tokens/RevShareNFT.sol
```diff
@@ -0,0 +1,281 @@
...
+    ) external onlyOwner nonReentrant mintChecks(pioneer) {
+        require(tokensToMint > 1, RevShareNFT__BatchMintMoreThanOne());
+
+        uint256 firstNewTokenId = totalSupply;
+        totalSupply += tokensToMint;
+        mintedToPioneers += tokensToMint;
+        uint256 lastNewTokenId = firstNewTokenId + tokensToMint;
+
+        // Update the revenues if the contract is set up to do so
+        _updateRevenuesIfProtocolIsSetUp(pioneer, operator);
```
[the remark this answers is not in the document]
maikelordaz: It is, that function is an external call

contracts/tokens/RevShareNFT.sol
```diff
@@ -0,0 +1,240 @@
...
+        returns (address revShareModule_)
+    {
+        // If the addressManager is not set, we return address(0)
+        if (address(addressManager) != address(0)) {
+            // We try to fetch the RevShareModule address from the AddressManager, if it fails it means
+            // the revenue share module is not set up, so we return address(0)
+            try addressManager.getProtocolAddressByName("REVENUE_SHARE_MODULE") {
+                revShareModule_ = addressManager
+                    .getProtocolAddressByName("REVENUE_SHARE_MODULE")
+                    .addr;
```
elallamTaka, line 221: Here getProtocolAddressByName is being called twice, is there a way to call it once?

contracts/tokens/RevShareNFT.sol
```diff
@@ -0,0 +1,240 @@
...
+        for (uint256 i = firstNewTokenId; i <= lastNewTokenId; ++i) {
+            if (!revUpdated) pioneerMintedAt[pioneer][i] = block.timestamp;
+            _safeMint(pioneer, i);
+        }
+
+        emit OnBatchRevShareNFTMinted(pioneer, firstNewTokenId, lastNewTokenId);
+    }
+
+    /**
+     * @notice Transfer override function. transfers are only allowed if the RevShareModule is set up
```
elallamTaka, line 156: this isnt an override

contracts/tokens/RevShareNFT.sol
```diff
@@ -0,0 +1,240 @@
...
+            _fetchRevShareModuleAddressIfIsSetUp() != address(0),
+            RevShareNFT__RevShareModuleNotSetUp()
+        );
+
+        // Update the revenues if the contract is set up to do so
+        _updateRevenuesIfProtocolIsSetUp(from);
+        _updateRevenuesIfProtocolIsSetUp(to);
+
+        super.transferFrom(from, to, tokenId);
+    }
```
elallamTaka, line 192: Why would a user use this over the transfer function above?

elallamTaka (changes requested):

contracts/tokens/RevShareNFT.sol
```diff
@@ -0,0 +1,240 @@
...
+    /*//////////////////////////////////////////////////////////////
+                                  MINT
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @notice Mint single token to a user
+     * @param pioneer The address of the member to mint a single NFT. This is only used for SINGLE_MINT
+     * @dev Only callable by someone with owner
+     */
+    function mint(address pioneer) external onlyOwner mintChecks(pioneer) {
```
elallamTaka, line 114: nonReentrant is used with batch mint but not in the mint or transfer

contracts/tokens/RevShareNFT.sol
```diff
@@ -0,0 +1,240 @@
...
+
+        super.transferFrom(from, to, tokenId);
+    }
+
+    /*//////////////////////////////////////////////////////////////
+                                GETTERS
+    //////////////////////////////////////////////////////////////*/
+
+    function tokenURI(uint256 tokenId) public view override returns (string memory) {
+        return
```
elallamTaka, line 199: Is it okay to return the URI for a non minted token since this does not prevent it?

elallamTaka (changes requested):

contracts/tokens/RevShareNFT.sol
maikelordaz: Commit 46f238daee1e2121567f0a5398e8d8ff6f90cea1

contracts/tokens/RevShareNFT.sol
maikelordaz: There are three scenarios.
1. We implement a functionality where we transfer some NFT from someone else. I dont know a benefit that you will pay only with the NFT and we handle the transfer using this function. Of course the user would need to set an approval. This would be a non malicious scenario.
2. A user gave approval for the token to an external malicious address without noticing, then this address would be able to transfer the token to himself. This would be a malicious transaction, actually nothing to do here after the transfers are allowed this would be completely under the user responsability.
3. A user gave approval to himself using another address, then it will be able to bypass the transfer. This would be a malicious transaction, actually nothing to do here after the transfers are allowed this would be completely under the user responsability.

The importance is to revert all transfers until the revenue is launched so we can calculate easier the corresponding revenue

contracts/tokens/RevShareNFT.sol
maikelordaz: Commit 76304e67cde3d6f6b5a0e4a98302a981a532be0d

elallamTaka (approved):

PR #132, opened 2025-08-28
feat: implement rev share contract (TS-1332)

maikelordaz:


contracts/modules/moduleUtils/TLDModuleImplementation.sol
elallamTaka, line 1: general feedback on the naming, I noticed this is the only file where the acronym TLD is mentioned in the file name. I recommend we remove it from the name and rename isTLDModule to isValidModule in the interface. This follows the vision for having multiple DAOs in the future deployed via the same implementation.

maikelordaz: Fair. Commint 2d50847

contracts/modules/RevShareModule.sol
```diff
@@ -0,0 +1,475 @@
...
+        contributionToken.safeTransferFrom(msg.sender, address(this), amount);
+
+        emit OnDeposit(amount);
+    }
+
+    /**
+     * @notice Sweep the non approved deposits to avoid accounting issues
+     * @dev Only callable by an operator
+     */
+    function sweepNonApprovedDeposits() external onlyRole(Roles.OPERATOR, address(addressManager)) {
```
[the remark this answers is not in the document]
maikelordaz: Yes, definitely. Actually I was thinking what to do with the non approved ones

contracts/modules/RevShareModule.sol
```diff
@@ -0,0 +1,475 @@
...
+    /**
+     * @notice Notify the contract about new revenue to be distributed
+     * @param amount The amount of revenue to be distributed
+     * @dev Only callable by specic contracts
+     * @dev The contract must have enough allowance to transfer the tokens
+     * @dev Splits deposits between Takadao (25%) and pioneers (75%)
+     * @dev Sets new reward rates for both pools, with carry-over if the previous stream was not finished
+     * @dev Increases the approved deposits for accounting purposes
+     */
+    function notifyNewRevenue(
```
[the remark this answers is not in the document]
maikelordaz: Yes. Non approved or non wanted deposits will cause in this case something similar to the vault inflation attack. Also could lead to issues with accounting, for example if we only have what comes from fees, and we now thats 10k, but we have in balance 20k, all the shares and rates will be wrongly distributed. To protect from the vault inflation attack there are four usual protections, minimum shares to avoid frontrunning, keep an internal balance to protect from donations, the contract is always the first depositor to avoid dead shares, and lately the ERC4626 by OpenZeppelin, ours should at least protect from donation

contracts/modules/moduleUtils/ModuleImplementation.sol
```diff
@@ -0,0 +1,48 @@
...
+pragma solidity 0.8.28;
+
+abstract contract ModuleImplementation {
+    IAddressManager internal addressManager;
+
+    string public moduleName;
+
+    modifier onlyContract(string memory name, address addressManagerAddress) {
+        require(
+            AddressAndStates._checkName(addressManagerAddress, name),
```
elallamTaka, line 23: _checkName takes in the address first, then the name, but the modifier takes in the name first, and then the address. is there a particular reason why the order is switched? I recommend to standardize the order across all functions and modifiers. It reduces the cognitive load on the reader.

contracts/modules/moduleUtils/ModuleImplementation.sol
```diff
@@ -0,0 +1,48 @@
...
+            ModuleErrors.Module__NotAuthorizedCaller()
+        );
+        _;
+    }
+
+    modifier onlyRole(bytes32 role, address addressManagerAddress) {
+        require(
+            AddressAndStates._checkRole(addressManagerAddress, role),
```
elallamTaka, line 31: Whats the benefit of this pattern here. `_checkRole` is used rather than directly calling `addressManger.hasRole` ? I noticed that in some modules, `hasRole` is being called directly.

contracts/modules/RevShareModule.sol
```diff
@@ -0,0 +1,493 @@
...
+    }
+
+    /*//////////////////////////////////////////////////////////////
+                                DEPOSITS
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @notice Notify the contract about new revenue to be distributed
+     * @param amount The amount of revenue to be distributed
+     * @dev Only callable by specic contracts
```
elallamTaka, line 154: small typo `specific`

maikelordaz: Commit 76cccd1f4cfb3d9c02858212eead793be1c903e4

contracts/modules/moduleUtils/ModuleImplementation.sol
maikelordaz: Commit 3e17cb608eea3b11e1421aa10f57855995b5d6f9

contracts/tokens/RevShareNFT.sol
```diff
@@ -137,36 +147,77 @@ contract RevShareNFT is
     ) external onlyOwner nonReentrant mintChecks(pioneer) {
         require(tokensToMint > 1, RevShareNFT__BatchMintMoreThanOne());
```
elallamTaka, line 148: the modifier mintChecks checks that the current supply is less than the total, but it doesnt take the `tokensToMint` into consideration. Say current is 8800, and batch mint is called with tokens to mint as 30, which will go above the max of 8820.

contracts/modules/moduleUtils/ModuleImplementation.sol
maikelordaz: Commit fdbde759727dcf09586598dd0cd7cb4d0fe64ca7

contracts/modules/RevShareModule.sol
```diff
@@ -0,0 +1,493 @@
...
+     * @param timestamp The timestamp when revenues will be available to claim
+     */
+    function setAvailableDate(
+        uint256 timestamp
+    ) external onlyRole(Roles.OPERATOR, address(addressManager)) {
+        AddressAndStates._onlyModuleState(
+            IAddressManager(addressManager).getProtocolAddressByName("MODULE_MANAGER").addr,
+            address(this),
+            ModuleState.Enabled
+        );
```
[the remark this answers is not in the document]
maikelordaz: Tried it but the compiler takes modifiers different, so they add a lot of size to the contract, and also it counts the stack different, and it adds easier a stack too deep error. Could be done, but in this PR breaks the compilation. Maybe in the PR for the architecture can be changed, as there most of the modules have a smaller size

PR #137, opened 2025-09-30
docs: arch design sequence diagram for subscribe refund use case (TS-2460)

maikelordaz:


docs/sequence_diagrams/Subscriptions/subscription-without-benefit.puml
```diff
@@ -0,0 +1,44 @@
...
+
+Backend -> SubscriptionModule: paySubscriptionOnBehalfOf(memberWallet, parentWallet, contribution)
+
+PRECHECKS_CALLER_MODULE()
+PRECHECKS_CONTRIB_MIN25()
+LOAD_MEMBER_AND_VALIDATE()
+OPTIONAL_PARENT_VALIDATE()
+
+' Asignar reward si hubo parent
+opt [there-was-a-parent]
```
elallamTaka: whats the different between adding - for these conditions and not adding dashshes?

docs/sequence_diagrams/Subscriptions/subscription-without-benefit.puml
```diff
@@ -0,0 +1,44 @@
...
+  SubscriptionModule -> RewardsModule: assignParentReward
+end
+
+REGISTER_AND_CONFIRM()
+
+CouponPool -> SubscriptionModule: transfer contribution
+KYC_OPT()
+
+opt [>30-days-passed-or-apply-to-service]
```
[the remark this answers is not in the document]
elallamTaka: I also suggest we separate the condition for the [apply for service]. And add a REF under a condition [> 30 days] then reference subscription with proration else subscription without proration. 

The idea is that the diagram for each use case must reference any related flow inside the same use case.

docs/sequence_diagrams/Subscriptions/refund.puml
```diff
@@ -0,0 +1,53 @@
...
+OPTIONAL_PARENT_VALIDATE()
+
+' Si hubo parent, asigna reward inicial
+opt [there-was-a-parent]
+  SubscriptionModule -> RewardsModule: assignParentReward
+end
+
+REGISTER_AND_CONFIRM()
+CouponPool -> SubscriptionModule: transfer contribution
+KYC_OPT()
```
elallamTaka: If you reference the Refund flow inside the main subscription flow, you will not need to repeat these steps. This makes the diagram shorter and easier to read.

docs/sequence_diagrams/Subscriptions/subscription-with-benefit-no-prorate.puml
```diff
@@ -0,0 +1,57 @@
...
+' Asignar reward por la contribución mínima (si hubo parent)
+opt [there-was-a-parent]
+  SubscriptionModule -> RewardsModule: assignParentReward(for the minimum contribution)
+end
+
+REGISTER_AND_CONFIRM()
+
+' Flujo de fondos/beneficios (no prorate)
+CouponPool -> SubscriptionModule: transfer minimum contribution
+KYC_MANDATORY()
```
elallamTaka: Same idea with the refund, by referencing this subflow, these do not require to be repeated anymore. The difference I notice is with the KYC_OPT vs Mandatory. Here I suggest there is revert inside join benefit if the user is not KYCed

docs/sequence_diagrams/Subscriptions/subscription-with-benefit-no-prorate.puml
```diff
@@ -0,0 +1,57 @@
...
+Backend -> LifeBenefitModule: joinBenefit(memberWallet)
+LifeBenefitModule -> MembersData: updateMember(memberWallet)
+
+SubscriptionModule -> Reserve: transfer minimum contribution to Reserve
+
+opt [there-was-a-parent]
+  SubscriptionModule -> RewardsModule: rewardParent
+  RewardsModule -> RewardsModule: Emits event
+  RewardsModule --> Parent: transfer parent reward
+end
```
elallamTaka: This is very repetitive, perhaps can be a macro too

docs/sequence_diagrams/Subscriptions/subscription-with-benefit-no-prorate.puml
```diff
@@ -0,0 +1,57 @@
...
+REGISTER_AND_CONFIRM()
+
+' Flujo de fondos/beneficios (no prorate)
+CouponPool -> SubscriptionModule: transfer minimum contribution
+KYC_MANDATORY()
+
+Backend -> LifeBenefitModule: joinBenefit(memberWallet)
+LifeBenefitModule -> MembersData: updateMember(memberWallet)
+
+SubscriptionModule -> Reserve: transfer minimum contribution to Reserve
```
elallamTaka: I can see that the call to joinBenefit invokes many actions, but its not clear from the diagram how the responses are handled. For example, say the transfer to reserve here has failed. Where is the failure communicated and to whom? (in this case I assume its the life benefit module).

docs/sequence_diagrams/Subscriptions/subscription-without-benefit.puml
```diff
@@ -0,0 +1,44 @@
...
+actor Backend
+actor Parent
+participant SubscriptionModule
+participant KYCModule
+participant MembersData
+participant RewardsModule
+participant CouponPool
+participant Reserve
+
+Backend -> SubscriptionModule: paySubscriptionOnBehalfOf(memberWallet, parentWallet, contribution)
```
elallamTaka: Just a 2 or 3 days back Mohammed told me that we might starting accepting different plans for the association. Ex. 25 vs 50 vs 250. Thus the subscription module needs a way to add a plan and validate it.

docs/sequence_diagrams/Subscriptions/subscription-with-benefit-no-prorate.puml
maikelordaz: Could not make it work as the caller sometimes is the SubscrtionModule and sometimes the LifeBenefitModule. But let me research how can I write it

docs/sequence_diagrams/Subscriptions/refund.puml
maikelordaz: Commit 8267a0d7e4c93dd4c44bffc82247f2f4c9d64d97

docs/sequence_diagrams/Subscriptions/subscription-without-benefit.puml
maikelordaz: Commit 606710593dbbaed54d70a6d71a3182b2fabffca2

maikelordaz:
@elallamTaka Have you reviewed the last changes in this PR?

Issue #141, opened 2025-10-01
[Info-02] redundant checks in `EnumerableSet` management


Brivan-26:
### Description
Protocol roles are managed using OpenZeppelin’s `EnumerableSet`. The `AddressManager` provides two functions for adding and removing roles:
```solidity 
function createNewRole(bytes32 newRole) external onlyOwner returns (bool success) {
    require(!_protocolRoles.contains(newRole), AddressManager__RoleAlreadyExists()); 
    success = _protocolRoles.add(newRole);
    emit OnRoleCreated(newRole);
}

function removeRole(bytes32 roleToRemove) external onlyOwner returns (bool success) {
    require(_protocolRoles.contains(roleToRemove), AddressManager__RoleDoesNotExist());
    success = _protocolRoles.remove(roleToRemove); 
    emit OnRoleRemoved(roleToRemove);
}
```
Both functions explicitly check the existence of the role before invoking `add` or `remove`. However, these checks are redundant because the underlying `EnumerableSet` implementation already performs existence validation. For example, the `add` function internally checks whether the value exists before insertion:
```solidity
// EnumerableSet.sol

function _add(Set storage set, bytes32 value) private returns (bool) {
    if (!_contains(set, value)) {
        set._values.push(value);
        // The value is stored at length-1, but we add 1 to all indexes
        // and use 0 as a sentinel value
        set._positions[value] = set._values.length;
        return true;
    } else {
        return false;
    }
}
```
Thus, the manual require statements add no functional security benefit and only increase code complexity.

### Recommendations
Consider removing the redundant require checks in createNewRole and removeRole. If explicit handling of role existence is desired, rely on the boolean return value of `_protocolRoles.add` and `_protocolRoles.remove` instead of duplicating checks.

PR #145, opened 2025-10-02
chore: remove check (TS-2526)

maikelordaz:
Related to https://github.com/TakafulDAO/takasure.sc/issues/141

### DIFF
--- a/contracts/managers/AddressManager.sol
+++ b/contracts/managers/AddressManager.sol
@@ -65,7 +65,6 @@ contract AddressManager is
     error AddressManager__AddressAlreadyExists();
     error AddressManager__AddressDoesNotExist();
     error AddressManager__AddModuleManagerFirst();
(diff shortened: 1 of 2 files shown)


contracts/managers/AddressManager.sol
```diff
@@ -184,7 +183,6 @@ contract AddressManager is
      * @return success A boolean indicating whether the role was successfully created
      */
     function createNewRole(bytes32 newRole) external onlyOwner returns (bool success) {
-        require(!_protocolRoles.contains(newRole), AddressManager__RoleAlreadyExists());
         success = _protocolRoles.add(newRole);
```
Brivan-26, line 186: If the role already exists, the `add` will silently return the `false` status, I'd prefer keeping it this way.  If explicit handling of role existence is desired, you can assert the `success` state.

Brivan-26 (approved):

Issue #139, opened 2025-10-01
[M-03] Inconsistent execution and partial DoS of `claimRevenueShare` after emergency withdraw


Brivan-26:
### Description
During an emergency withdraw, the contract resets `approvedDeposits` to zero:
```solidity 
function emergencyWithdraw() external onlyRole(Roles.OPERATOR, address(addressManager)) {
    IERC20 contributionToken = IERC20(
        addressManager.getProtocolAddressByName("CONTRIBUTION_TOKEN").addr
    );
    uint256 balance = contributionToken.balanceOf(address(this));

    // Halt streams & reset accounting
    rewardRatePioneersScaled = 0;
    rewardRateTakadaoScaled = 0;
    periodFinish = block.timestamp;
    lastUpdateTime = block.timestamp;
@==>    approvedDeposits = 0; // Reset approved deposits

    contributionToken.safeTransfer(msg.sender, balance);
}
```
This creates an inconsistent execution flow in `claimRevenueShare`, as the function requires `approvedDeposits` to always be greater than the revenue being withdrawn:
```solidity 
function claimRevenueShare() external nonReentrant returns (uint256 revenue) {
    // --SNIP
    if (revenue > 0) {
        revenuePerAccount[beneficiary] = 0;

        IERC20 contributionToken = IERC20(
            addressManager.getProtocolAddressByName("CONTRIBUTION_TOKEN").addr
        );
        contributionToken.safeTransfer(beneficiary, revenue);

@==>        if (approvedDeposits < revenue) revert RevShareModule__InsufficientApprovedDeposits();
        approvedDeposits -= revenue;

        emit OnRevenueShareClaimed(beneficiary, revenue);
    }
}
```
Since `approvedDeposits` is forcibly set to zero during an emergency withdraw, all users with unclaimed revenue are permanently blocked from withdrawing their accrued rewards.

Attempts to work around this issue are ineffective:
1. Directly funding the contract with the required tokens does not work, as the `approvedDeposits` check will still revert.
2. Calling `notifyNewRevenue` to increase `approvedDeposits` is also invalid. This updates `lastUpdateTime`, causing the new funds to be treated as fresh revenue for redistribution. Even if this temporarily resolves withdrawals, it leads to inconsistent distribution of future rewards, preventing certain users from receiving their entitled revenue.

## Recommendations
[...]

PR #147, opened 2025-10-02
revise: fix partiao DoS option 1 (TS-2529)

maikelordaz:
Related to https://github.com/TakafulDAO/takasure.sc/issues/139


maikelordaz:
> emergency mode should also prevent notifyNewRevenue from being called since it means there is something wrong

Fixed in commit 9b3aa83c725a17fc4fb14ecc6d5af72eaa1f0d70

PR #149, opened 2025-10-08
revise: subscription flow (TS-2521)

maikelordaz:
This PR has:
- The subscription flow with the new architecture
- A new module to use as protocol global storage, only for vars that could be needed between modules

To have in mind:
To make it easier to review I remove the other modules that still need to be written like this, so I also removed the suscription related tests because I have them already written. I dont consider necesary to write them now to modify them completely in next PR.

Next one will be the `KYCModule` and `ReferralRewardsModule` with those ones the tests can be rewritten to check the complete interaction


contracts/modules/SubscriptionModule.sol
```diff
@@ -3,569 +3,394 @@
...
-            couponAmount
-        );
+        _paySubscription(userWallet, parentWallet, couponAmount, membershipStartTime);
 
         if (couponAmount > 0) {
-            isMemberCouponRedeemer[memberWallet] = true;
-            emit TakasureEvents.OnCouponRedeemed(memberWallet, couponAmount);
+            isMemberCouponSubscriptionRedeemer[userWallet] = true;
+            emit TakasureEvents.OnCouponRedeemed(userWallet, couponAmount);
         }
```
elallamTaka: 1. I don't think its necessary to emit an event for coupon redemptions specifically. Rather, emit one event when a subscription is paid, and this would be the event that the backend will process. It can contain the coupon amount if any. 

2. I have a suggestion with regards to the coupon redeemer mapping. I noticed its used for refunds specifically. How about we add `couponAmountRedeemed` as a property of an association member that resets to 0 after refund? This way we get rid of the private mapping. Are you defining it as private mapping since other modules will not be required to read it? And are you defining `isRefunded` as a property in the member because it will be read by other modules? This will help me understand the convention you choose for vars.

contracts/modules/ProtocolStorageModule.sol
```diff
@@ -0,0 +1,318 @@
...
+        bool isRejoin
+    ) internal view {
+        if (!isRejoin)
+            require(
+                members[_member.wallet].wallet == address(0),
+                ModuleErrors.Module__AlreadyJoined()
+            );
+
+        require(_member.wallet != address(0), ModuleErrors.Module__InvalidAddress());
+        require(_member.wallet != _member.parent, ModuleErrors.Module__InvalidAddress());
```
elallamTaka: Since these are giving out the same error why not combine them into one require

contracts/modules/ProtocolStorageModule.sol
```diff
@@ -0,0 +1,318 @@
...
+
+        // The user state must be inactive or canceled and not refunded
+        require(
+            (_member.memberState == AssociationMemberState.Inactive ||
+                _member.memberState == AssociationMemberState.Canceled) && !_member.isRefunded,
```
elallamTaka: But a refunded user can still rejoin?

contracts/modules/ProtocolStorageModule.sol
```diff
@@ -0,0 +1,318 @@
...
+        AssociationMember memory member
+    ) external onlyProtocolOrModule {
+        // The module must be enabled
+        AddressAndStates._onlyModuleState(
+            ModuleState.Enabled,
+            address(this),
+            addressManager.getProtocolAddressByName("MODULE_MANAGER").addr
+        );
+        _associationMemberProfileChecks(member, true);
+        require(members[member.wallet].wallet != address(0), ModuleErrors.Module__InvalidAddress());
```
elallamTaka: I recommend you add this require inside `_associationMemberProfileChecks` since you already have a check `if(!isRejoin)` ...

```
if (isRejoin)
    require(members[member.wallet].wallet != address(0), ModuleErrors.Module__InvalidAddress());
else require( members[_member.wallet].wallet == address(0), ModuleErrors.Module__AlreadyJoined());
```

contracts/modules/SubscriptionModule.sol
```diff
@@ -3,569 +3,394 @@
...
     /**
      * @notice Method to refunds a user
-     * @dev To be called by anyone
+     * @dev To be called by the operator only
      * @param memberWallet address to be refunded
      */
-    function refund(address memberWallet) external {
+    function refund(
+        address memberWallet
+    ) external onlyRole(Roles.OPERATOR, address(addressManager)) {
```
elallamTaka: I dont think this is for the operator to call (which is a multsig), it should be the coupon redeember, or we can add another role called `RefundAdmin` for the backend

contracts/modules/SubscriptionModule.sol
```diff
@@ -3,569 +3,394 @@
...
      * @notice All refunds for users that used coupons will be restored in the coupon pool
      *         The user will need to reach custommer support to get the corresponding amount
      */
     function _refund(address _memberWallet) internal {
+        // Module must be enabled
         AddressAndStates._onlyModuleState(
             ModuleState.Enabled,
             address(this),
-            IAddressManager(addressManager).getProtocolAddressByName("MODULE_MANAGER").addr
+            addressManager.getProtocolAddressByName("MODULE_MANAGER").addr
```
elallamTaka: Here you are checking for the module state of the subscribe module, but you dont have this check for the other calls. I recommend we stick to having this type of check inside external functions only for consistency.

contracts/modules/SubscriptionModule.sol
```diff
@@ -3,569 +3,394 @@
...
+        if (msg.sender == lifeBenefitModuleAddress || msg.sender == farewellBenefitModuleAddress) {
+            // If the caller is a benefit then we consider this as a contribution
+            revenueType = RevenueType.Contribution;
+        } else if (isOperator) {
+            // If the caller is the operator, then we consider this as a donation, and have to be called after the refund period ends
             require(
-                newMember_.memberId != 0 && newMember_.wallet == _memberWallet,
-                ModuleErrors.Module__WrongMemberState()
+                block.timestamp >= member.associateStartTime + 30 days,
+                ModuleErrors.Module__InvalidDate()
```
elallamTaka: This is not true because members can opt to donate before 30 days to get access to referral link

contracts/modules/SubscriptionModule.sol
```diff
@@ -3,569 +3,394 @@
...
+                           INTERNAL FUNCTIONS
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @notice Allow new members to pay subscriptions. All members must pay first, and KYC afterwards. Prejoiners are KYCed by default.
+     * @param _userWallet address of the member
+     * @param _parentWallet address of the parent
+     * @param _couponAmount amount of USDC in six decimals to be used as a coupon
+     */
+    function _paySubscription(
```
elallamTaka: The logic for recurring payments is missing from this module. This can effect how the structs look like.

elallamTaka (changes requested):
Overall, clear, straightforward, very modular practices. Well done!

contracts/modules/SubscriptionModule.sol
```diff
@@ -3,569 +3,394 @@
...
+            _transferFee(contributionToken, couponPool, _fee);
         }
-        return (_reserve);
+
+        // Transfer the referral reserve amount to the corresponding module
+        if (_toReferralReserveAmount > 0)
+            contributionToken.safeTransfer(_referralRewardsModule, _toReferralReserveAmount);
+    }
+
+    function _transferFee(IERC20 _contributionToken, address _userWallet, uint256 _fee) internal {
```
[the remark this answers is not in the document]
maikelordaz: Commit fa3e78e379712d79f22e63f98f7d34177c5ab378

contracts/modules/SubscriptionModule.sol
maikelordaz: Commit e25de7180c9fb6e9b9c7a28d67ec4484f1bdb7b8

contracts/modules/SubscriptionModule.sol
maikelordaz: Commit c04ae375e0d1514a0e98d93f7a14c8cab29fbca4

contracts/modules/SubscriptionModule.sol
maikelordaz: Discussed on architecture call

contracts/types/TakasureTypes.sol
```diff
@@ -67,7 +60,22 @@ struct PrepaidMember {
...
+struct AssociationMember {
+    uint256 memberId;
+    uint256 discount;
+    uint256 associateStartTime; // in seconds
+    address wallet;
+    address parent;
+    AssociationMemberState memberState;
+    bool isRefunded; // Can not be true if isKYCVerified is true
+    bool isLifeProtected;
+    bool isFarewellProtected;
```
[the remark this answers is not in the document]
maikelordaz: Used an array for two reasons:
1. Is not possible to loop through mapping without knowing the keys
2. An struct with nested mappings can not be an input in contract external calls, only input in internal calls and external calls in libraries. This makes impossible to call `ProtocolStorageModule::createAssociationMember` or `ProotocolStorageModule::updateAssociationMember`

elallamTaka (approved):

PR #188, opened 2026-03-30
revise: use different swap pools

maikelordaz:


MohammedTaka (approved):
Looks good, would prefer more flexibility in the choosing of the pools but all in all this is fine.

PR #191, opened 2026-04-26
revise: implement router and token smart contracts (TS-3336)

maikelordaz:


contracts/helpers/libraries/constants/StorageKeys.sol
elallamTaka: What are these used for and whats the design pattern?

contracts/modules/KYCModule.sol
```diff
@@ -61,7 +62,13 @@ contract KYCModule is ModuleImplementation, Initializable, UUPSUpgradeable, Reen
      * @dev It reverts if the member is the zero address
      * @dev It reverts if the member is already KYCed
      */
-    function approveKYC(address memberWallet) external onlyRole(Roles.KYC_PROVIDER, address(addressManager)) {
+    function approveKYC(address memberWallet) external {
+        require(
+            addressManager.hasRole(Roles.KYC_PROVIDER, msg.sender)
+                || addressManager.hasName("PROTOCOL__ROUTER", msg.sender),
```
elallamTaka: can we use has role here for the router as well, its safer to stick with one.

contracts/managers/TLDRouter.sol
```diff
@@ -0,0 +1,526 @@
...
+
+        // Create/update subscription state first and settle the returned USDC plan atomically.
+        // If either module execution or settlement fails, the whole command reverts.
+        plan_ = _dispatchSettlementCommand(TLDCommands.SUBSCRIBE, _inputs, _caller);
+
+        // After subscription settlement succeeds, mint the membership TAKA allocation.
+        // TakaLockModule handles first-join locking vs renewal transfer behavior.
+        uint256 allocation = _takaAllocation(planPrice);
+        ITakaLockModule(_addressByName("MODULE__TAKA_LOCK")).purchaseAndLock(member, allocation, planPrice);
+    }
```
elallamTaka: This pattern gives the router business responsibility which should not be the case. Router must only router, and move money. The biz checks, including caller must be defined in the modules themselves. The subscribe module must talk to the TakaLock module directly to purchase. Thus if biz logic changes for any reason, modules must be upgraded, not router.

contracts/modules/SubscriptionModule.sol
```diff
@@ -286,21 +313,29 @@ contract SubscriptionModule is
...
-     * @notice Allow new members to pay subscriptions. All members must pay first, and KYC afterwards. Prejoiners are KYCed by default.
+     * @notice Create or update member subscription state and return the router settlement plan
+     * @param _params Decoded subscription inputs
+     * @param preview True to calculate without mutating storage or referral accounting
+     * @return plan_ Settlement plan for the router to execute
+     * @dev All members pay first and KYC afterwards. Prejoiners are KYCed by default in legacy flows.
+     * @custom:invariants State-changing subscriptions must not transfer contribution tokens directly.
      */
-    function _paySubscription(PaySubscriptionParams memory _params) internal nonReentrant {
+    function _paySubscription(PaySubscriptionParams memory _params, bool preview)
```
elallamTaka: Now that the router can be called by both backend and member, this can be exploited to pay for "free" subscriptions. Attacker can make the call with a couponAMount and settle pulls that amount from the coupon pool. A check must be added, coupons can only be used by backend admin. Same for recurring.

elallamTaka (changes requested):

contracts/helpers/libraries/constants/TLDCommands.sol
```diff
@@ -0,0 +1,49 @@
...
+library TLDCommands {
+    /*
+    * From 0x01 to 0x14 are reserved for subscription management module
+    * From 0x15 to 0x28 are reserved for benefit modules
+    * From 0x29 to 0x3c are reserved for token lock module
+    * From 0x3D to 0x50 are reserved for saving vault
+    * From 0x51 tp 0x64 are reserved for card modules
+    * From 0x65 to 0x78 are reserved for reward/admins
+    * From 0x79 to 0x7f are reserved for future use
+    * Utility
```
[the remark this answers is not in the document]
maikelordaz: Commit 4fdd27f6fa67eb30f7ff069ed2318c205742b14c

contracts/helpers/libraries/constants/StorageKeys.sol
maikelordaz: `StorageKeys.sol` is like a canonical place for string keys used with the generic key/value storage helpers in `MainStorageModule`.

The idea is:

- modules should not repeat string literals like `"benefitFee"` or `"allowUserInitiatedCalls"`
- I put shared keys in one library
- the key name is human-readable, the storage slot is deterministic.

contracts/modules/TakaLockModule.sol
```diff
@@ -0,0 +1,527 @@
...
+     * @param usdcAllocation USDC-equivalent allocation used to calculate minted TAKA
+     * @param planPrice Subscription plan price used to derive Basic or Premium tier
+     * @dev Router-only. First-time members lock one TAKA; renewals receive the full minted amount.
+     * @custom:invariants First-time active members receive exactly one TAKA locked when enough TAKA is minted.
+     * @custom:invariants Renewals do not increase the automatic membership lock.
+     */
+    function purchaseAndLock(address member, uint256 usdcAllocation, uint256 planPrice)
+        external
+        nonReentrant
+        onlyRouter(address(addressManager))
```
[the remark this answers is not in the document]
maikelordaz: Commit 260b3cfb857e51ccc58c76732d58f7da1781b8d2

contracts/modules/SubscriptionModule.sol
```diff
@@ -316,143 +394,375 @@ contract SubscriptionModule is
...
-        // The member should exist and not be refunded
+        // The member must exist and must not have been refunded before.
         require(_member.wallet == _memberWallet, ModuleErrors.Module__InvalidAddress());
         require(!_member.isRefunded, SubscriptionModule__NothingToRefund());
         uint256 currentTimestamp = block.timestamp;
         uint256 startTime = _member.associateStartTime;
 
-        // The member can refund before 30 days of the payment
+        // Refunds are only allowed during the first 30 days after association start.
         uint256 limitTimestamp = startTime + (30 days);
```
elallamTaka: 30 days in not true anymore, its 14 days. Best make it configurable.

elallamTaka (changes requested):
To-do: 
- [ ] Add cashback logic. Plan must contain cashback amount, if amount is > 0 then that amount must be transferred from "cashback vault" to users address. Only transfer on first payment, not on recurring. 

Note, we agreed that modules must orchestrate business. But we also agreed in our design pattern that the router orchestrates/ executes settlements. However, some business logic requires settlements. To maintain our design pattern, we can have the settlement plan return multiple legs, not only one. Take purchase and lock for example: 
Router --> Subscription --> return plan with two legs, one to pay contribution and one to purchase Taka. This way settlement does not happen inside subscription anymore, but the logic is determined by subscriptions module. Any complications you can think of? 

Purchase taka can happen through an exchange in the future, contrary to how the current PR simply mints.

PR #192, opened 2026-04-26
feat: add BTC-ETH investment vault strategy (IFBtcEthStrategy)

MohammedTaka:
# feat: add BTC-ETH investment vault strategy (`IFBtcEthStrategy`)

## Summary

Adds a new investment strategy for a Uniswap V3 WBTC/WETH LP position, plus the helpers and library it depends on. Lives under a new `contracts/investFunds/` tree so it's clearly separated from the existing SaveFunds strategies.

The strategy is a sub-strategy of the existing `SFStrategyAggregator` (no aggregator/vault changes so these contracts can be reused). USDC is the underlying. The contract operates in two modes:

- **HODL** — idle WBTC + WETH reserves, no active position
- **LP** — concentrated V3 position in the WBTC/WETH 0.05% pool with a configurable tick range

Off-chain control flow: a Chainlink CRE workflow runs the correlation algorithm every 12h / 24h, decides between `ENTER_LP` / `EXIT_LP` / `RECENTER` / hold, and (when needed) submits a pre-built rebalance payload through the aggregator. Withdrawals are fully autonomous on-chain — no CRE involvement, just Chainlink price feeds for direction.

## What's in the branch

| File | Lines | Role |
|---|---|---|
| `contracts/investFunds/protocol/IFBtcEthStrategy.sol` | 1377 | Main strategy. UUPS proxy. Implements `ISFStrategy` + `ISFStrategyMaintenance`. |
| `contracts/investFunds/libraries/BtcEthDecimalMath.sol` | 194 | Pure math library for the 8 / 18 / 6 decimal system (WBTC / WETH / USDC). Built on Uniswap `FullMath`. |
| `contracts/investFunds/helpers/BtcEthSwapHelper.sol` | 246 | Universal Router execution helper. Reached via `delegatecall`. V3 + V4 single-hop builders + a generic `executePreparedSwap` for CRE-built calldata. |
| `contracts/investFunds/helpers/BtcEthLPHelper.sol` | 315 | LP lifecycle helper. Reached via `delegatecall`. Owns `mintPosition`, `increaseLiquidity`, `decreasePartialAndCollect`, `burnPositionFully`, `collectFees`. |
| `test/unit/invest-funds/btc-eth-strategy/BtcEthDecimalMathTest.t.sol` | — | 25 unit + fuzz tests for every decimal conversion path. |

Bytecode (Arbitrum, optimizer 450 runs):

```
IFBtcEthStrategy   24,434 B  /  24,576 limit  (142 B margin)
BtcEthLPHelper      5,847 B
BtcEthSwapHelper    4,845 B
BtcEthDecimalMath      85 B
```

### Asset model

The vault accepts USDC (6 dec) but the strategy holds it in three forms:

```
USDC (6)  ─── deposit-sweep ───►  WBTC (8) + WETH (18)  ─── enterLP ───►  V3 LP position
                                          ▲                                       │
                                          │                                       │
                                          └──────── exitLP / partial burn ────────┘
```

- **Underlying** (`asset()`): USDC.
- **HODL reserves**: idle WBTC + WETH balances on the strategy contract.
- **LP position**: a single WBTC/WETH NFT owned by the SFVault/IFVault (strategy holds operator approval).

### `totalAssets()` — pool TWAP, two-hop

Returns the strategy's total value in USDC, used by the aggregator/vault for share pricing on every deposit and withdraw. Costs ~145k gas vs ~700k for an equivalent Quoter call.

```
WBTC ─── lpPool TWAP (WBTC/WETH 0.05%)  ───►  WETH
                                                ▼
                                         wethUsdcPool TWAP
                                                ▼
                                              USDC
```

Both hops use `pool.observe()` over a configurable `twapWindow` (default 30 min) with a `slot0()` fallback if observe fails. WBTC/USDC is a Uniswap V4 pool, which doesn't expose per-pool `observe()`, so we route around it with two V3 TWAPs. Chainlink is **not** used for valuation — only for withdraw direction decisions, where rough USD prices are fine.

### Deposit flow (CRE-directed)

The CRE deposit-sweep workflow (runs every 12-24h) reads `vault.idleAssets()`, checks Chainlink for the underweight HODL token, and submits:

```
vault.investIntoStrategy(amount, [strategy], [DepositData])
   └─► aggregator.deposit(amount, bundle)
         └─► strategy.deposit(amount, payload)
```

`DepositData` carries USDC split per token, per-direction min outs, and an optional `addToLp` flag. The contract executes 0/1/2 swaps via the SwapHelper (USDC→WBTC via V4, USDC→WETH via V3) and optionally adds the resulting tokens to the LP position.

### Withdraw flow (autonomous, on-demand)

User → `vault.withdraw()` → `_ensureLiquidity()` → if shortfall, `aggregator.withdraw(shortfall, vault, bytes(""))` → aggregator falls through to the operator-set `defaultWithdrawPayload` → `strategy.withdraw(amount, aggregator, defaultPayload)`.

Inside the strategy:

1. Idle USDC may already cover the request —> done.
2. Otherwise compute `targetUsdc = shortfall * (1 + withdrawMarginBPS)`.
3. **If in LP mode**, check HODL value via Chainlink **before any swap**. If HODL is insufficient, partial-burn the LP first. This avoids the 4-swap scenario (HODL swap → still short → LP burn → 2 more swaps).
4. Determine overweight side via Chainlink, swap it for USDC, then swap the other side if still short.
5. **Strict guarantee**: if final USDC < requested, revert `LiquidationShortfall(delivered, requested)`. The request was already capped to `totalAssets()` up-front, so an undershoot indicates TWAP/execution-price divergence or slippage > margin — fail loud rather than silently shortchange the user.

Max 2 swaps in every path. The withdraw payload is just a deadline — pool route configs (V4 fee/tickSpacing/hooks for WBTC/USDC, V3 fee for WETH/USDC) live as strategy state, set once by the operator at deploy.

### Rebalance flow (CRE-driven, tagged-wrapper payload)

```solidity
abi.encode(uint8 actionType, bytes inner)
```

| Type | Action | Inner struct |
|---|---|---|
| 1 | `ENTER_LP` (HODL → LP) | `RebalanceEnterLpData` — tick range, LP amounts, optional WBTC↔WETH ratio swap, slippage floors |
| 2 | `EXIT_LP` (LP → HODL) | `RebalanceExitLpData` — burn floors, optional 50/50 swap |
| 3 | `RECENTER` | `RebalanceRecenterData` — burn old, swap to new ratio, mint at new ticks |
| 4 | `HODL_DRIFT` | `RebalanceSwap` — pure WBTC↔WETH swap when one side > 70% by USD value (per the lazy 50/50 model) |

All swap calldata inside these payloads is **pre-built by the CRE** — the contract decodes, validates, and forwards it to the Universal Router via `executePreparedSwap`. No on-chain route selection or Quoter calls.

### Aggregator compatibility

`SFStrategyAggregator._isUniV3Strategy()` probes for `pool()` and `otherToken()` getters to decide whether to apply the V3-specific withdraw payload schema. We deliberately **do not** expose those names — our public getters are `lpPool`, `wethUsdcPool`, `wbtc`, `weth`. This way the aggregator doesn't misclassify us and the new payload schema goes through correctly.

### Helpers via `delegatecall`

Both helpers are stateless contracts (only an `addressManager` immutable). Strategy reaches them via `delegatecall`, which means:

- `address(this)` inside the helper resolves to the strategy proxy
- Approvals, NFT ownership, and Permit2 state all live in strategy storage
- A guard `require(address(this) == _strategyAddress(), NotStrategyContext())` rejects direct calls

Events the helpers emit (`OnPositionMinted`, `OnLiquidityIncreased/Decreased`, `OnPositionCollected`, `OnSwapExecuted`) fire from the strategy proxy address. The strategy ABI must include those signatures so off-chain indexers can decode them.

## Decimal math

Three decimals (8 / 18 / 6) and conversions between every pair via either Chainlink price or pool sqrtPriceX96. Easy to get wrong silently. To contain the risk:

- **Dedicated library**, `BtcEthDecimalMath`, with one named function per direction. No inline scaling anywhere in the strategy.
- All multiplications go through Uniswap `FullMath.mulDiv` (512-bit intermediate) — no phantom overflow.
- The sqrtPriceX96 helpers (`token0AsToken1`, `token1AsToken0`) split at `uint128.max` to handle the high-price branch the same way Uniswap's `OracleLibrary.getQuoteAtTick` does.
- Fuzz coverage: every conversion path, round-trip consistency (within 1 unit at the 8-decimal floor), overflow boundaries, monotonicity, zero-amount handling. 25 cases, all passing.

## Configuration / deploy notes

The strategy needs four AddressManager entries to be wired up at deploy:

| Name | Resolves to |
|---|---|
| `PROTOCOL__SF_AGGREGATOR` |  `SFStrategyAggregator` |
| `PROTOCOL__IF_BTC_ETH_STRATEGY` | strategy proxy |
| `HELPER__BTC_ETH_SWAP` | `BtcEthSwapHelper` deployed instance |
| `HELPER__BTC_ETH_LP` | `BtcEthLPHelper` deployed instance |

After deploy, the operator must call:

```
strategy.setWbtcUsdcV4Pool(fee, tickSpacing, hooks)   // V4 pool key for WBTC/USDC swaps
strategy.setWethUsdcV3Fee(fee)                        // V3 fee tier for WETH/USDC swaps
aggregator.setDefaultWithdrawPayload(strategy, abi.encode(WithdrawData({ swapDeadline: 0 })))
```

The strategy must also be added to the aggregator's substrategy set (existing `addSubStrategy` flow) and the vault must approve the strategy for PositionManager NFTs (`positionManager.setApprovalForAll(strategy, true)`).

## What's deliberately NOT in this PR

- **Chainlink CRE workflow** (the off-chain TS algorithm + payload builder).
- **Auto-compound on harvest.** v1 keeps fees as WBTC/WETH HODL reserves, accounted in `totalAssets()` automatically. The `harvest(bytes data)` interface already accepts a payload so a future version can add compound instructions without an ABI change. We assume that deposits and withdrawals will be frequent enough that we don't have to accomodate for this.
- **No changes to `SFVault`, `SFStrategyAggregator`, or `SFUniswapV3Strategy`.** This is purely additive.

## Bytecode budget

`IFBtcEthStrategy` sits at 24,434 B with 142 B headroom against the 24KB Arbitrum limit. The two delegatecall helpers (`BtcEthSwapHelper`, `BtcEthLPHelper`) and the external `BtcEthDecimalMath` library exist for this reason — without them the strategy would be ~5KB over the limit.

## Test plan

- [x] `forge test --match-contract BtcEthDecimalMathTest` — 25 / 25 pass
- [x] `forge build --sizes` — IFBtcEthStrategy at 24,434 B (under limit)
- [ ] Testnet deployment with custom tokens and custom uniswap pools.
- [ ] Fork tests on Arbitrum mainnet (real pool state, real Chainlink + TWAP responses) — follow-up PR
- [ ] Integration tests for the full withdrawal chain (user → vault → aggregator → strategy) — follow-up PR
- [ ] V3 / V4 mixed swap tests via the SwapHelper — follow-up PR
- [ ] Slippage / MEV scenarios — follow-up PR

The follow-up PR will cover the integration / fork / scenario tests once the algorithm and CRE wiring side stabilizes.


MohammedTaka:
[eth-btc-pr (codex).md](https://github.com/user-attachments/files/27099685/eth-btc-pr.codex.md)
[eth-btc-pr (devin).md](https://github.com/user-attachments/files/27099686/eth-btc-pr.devin.md)

Feedback from Devin & Codex

MohammedTaka:
Summary of the fixes from claude:

# Review response — commit [`69ff907`](https://github.com/TakafulDAO/takasure.sc/pull/192/commits/69ff907)

Two independent reviews (Devin + Codex) raised 7 actionable items. All fixed in this commit. The remaining flags were considered and intentionally not changed — reasons inline below.

## Fixed

| # | Source | Issue | Fix |
|---|---|---|---|
| 1 | Devin Bug 1 | `_swapWbtcForUsdc` / `_swapWethForUsdc` derived `minOut` from `_targetUsdc` even after `_wbtcNeeded` was capped to balance. With a small balance the swap reverted and the fallback leg never ran. | Compute `_expectedOut = wbtcAsUsdc(_wbtcNeeded, ...)` AFTER the balance cap, then `_minOut = _expectedOut * (1 - swapSlippageBPS)`. Same fix applied symmetrically to the WETH variant. |
| 2 | Devin Bug 2 + Codex Medium | `BtcEthLPHelper._approve` skipped when `_amount == 0`. Reset calls after `mint` / `increaseLiquidity` were silent no-ops, leaving residual allowances on the strategy proxy. | Removed the `if (_amount > 0)` guard. `_approve` now always `forceApprove`s, including the zero reset. |
| 3 | Codex High | `RebalanceSwap` carried raw `(commands, inputs)` bytes and forwarded them to the Universal Router. A compromised keeper could craft a payload that exfiltrated funds via UR `SWEEP` / `TRANSFER` actions while setting `minOut = 0`. | Replaced with constrained struct `(bool sellWbtc, uint256 amountIn, uint256 minOut)`. The strategy now builds the V3 single-hop calldata on-chain via the existing `BtcEthSwapHelper.buildAndExecV3Swap`. `minOut > 0` is enforced when `amountIn > 0`. The dead `executePreparedSwap` path was removed from `BtcEthSwapHelper` (-599 B). |
| 4 | Codex High/Medium | `_burnLpForShortfall` called `_decreaseLiquidityAndCollect(..., 0, 0)`. A spot-price manipulation around a user-triggered withdraw could force a bad token mix with zero protection. | Min floors are now derived from TWAP-expected pro-rata amounts: `expected = posAmount * liqToBurn / liq`, then `min = expected * (1 - swapSlippageBPS)`. Mapped to (token0, token1) order. If actual decrease diverges from TWAP by more than `swapSlippageBPS`, decrease reverts and `LiquidationShortfall` fires. |
| 5 | Devin Investigate 1 | `deposit` returned `investedAssets_ = _assets` even when residual USDC was swept back to the vault. Aggregator overstated the invested amount. | `investedAssets_` is now computed from `_assets - residualUsdc` before the sweep — the aggregator gets the truth. |
| 6 | Codex Low | Chainlink validation only checked positive answer + `updatedAt <= 1h`. | Added `updatedAt != 0` check (catches uninitialised feed) and `answeredInRound >= roundId` check (catches stale-round attacks). Decimals stay hardcoded (single-chain v1 — both feeds are 8-decimal on Arbitrum). |

## Documented (not coded)

| # | Source | Reason |
|---|---|---|
| 7 | Codex Medium — TWAP→spot fallback | The fallback is intentional: reverting `_twapSqrtPrice` would brick `totalAssets()`, which would brick every `deposit` / `withdraw` / `previewDeposit` / `previewMint` the SFVault makes. Same pattern as the existing `SFUniswapV3Strategy`. Added explicit NatSpec describing the operational requirement (pre-warm observation cardinality before deploy) and the off-chain monitoring expectation (compare on-chain `totalAssets` against an external pricing source, alert on persistent divergence). The function is `view` so emitting a `OnTwapFallback` event isn't possible. |

## Considered, intentionally not changed

| Source | Item | Reason |
|---|---|---|
| Devin Flag 1 | delegatecall storage layout | Devin's analysis confirms the immutable-only design is correct. No action. |
| Devin Flag 2 | `burnPositionFully` skips collect when `liquidity == 0` | All in-strategy flows guarantee fees + liquidity are collected together. Adding a defensive `collect()` on the zero-liquidity branch would charge gas every call to guard against a state we don't create. |
| Devin Flag 3 | Permit2 max approval permanent | Documented gas optimisation; standard DeFi pattern; identical trust model to the existing `SFUniswapV3Strategy`. |
| Devin Flag 4 | Hardcoded Permit2 address | Canonical CREATE2 deployment on every EVM chain we'd target. Mutability isn't worth it for v1 single-chain Arbitrum. Worth revisiting if/when we go multi-chain. |
| Devin Flag 5 | TWAP rounding | Devin confirms it matches Uniswap's `OracleLibrary` convention. |
| Devin Flag 6 | Stale `tokensOwed` undercount | Universal Uniswap V3 lazy-fee-accounting limitation — same in `SFUniswapV3Strategy`. Fixing requires the `PositionValue.fees()` math (~2 KB more bytecode against a ~230 B margin). The undercount is bounded by `recentVolume × feeTier`; negligible for valuation purposes. |
| Codex Low | V4 `amountOutMin` uint128 cast in `SFUniswapV3SwapRouterHelper` | Pre-existing SaveFunds helper, not modified by this PR. Out of scope. |

## Bytecode impact

| Contract | Before | After | Δ |
|---|---|---|---|
| `IFBtcEthStrategy` | 24,434 (142 B margin) | **24,343 (233 B margin)** | **-91** |
| `BtcEthSwapHelper` | 4,845 | 4,246 | -599 |
| `BtcEthLPHelper` | 5,847 | 5,841 | -6 |

The constrained-rebalance refactor (#3) was net-positive for headroom because removing `executePreparedSwap` from `BtcEthSwapHelper` saved more bytecode than the new validation in `_execRebalanceSwap` cost.

## Tests

- `forge test --match-contract BtcEthDecimalMathTest`: 25 / 25 pass
- `forge build --sizes`: clean

Off-chain encoder/decoder tests in the consumer repo will need to be updated to match the new `RebalanceSwap` shape `(bool, uint256, uint256)` — that work is tracked in the consumer-side branch.

